PR #10, opened 2017-01-01
Update to MC 1.11.2

MitchDevelopment:
Updated to MC 1.11.2
Removed you're maven repo as it wasn't needed.
Fixed .gitignore so that it ignores the .idea/ directory and *.iml file.


aikar:
Sorry I want all of that there.

My repo is in anticipation if any other libs are needed, and more reliable for the project to build.

Issue #3, opened 2016-12-22
Auto Help System


aikar:
Provide a @Description type annotation for documenting purpose of command

Provide mechanisms for automatic help display, such as failed to find subcommands

aikar:
Looking for feedback from community on https://github.com/aikar/commands/wiki/Command-Help before we mark this "Stable" and done.

PR #18, opened 2017-04-27
Add Paginator Functional Interface and ItemPaginator class

simpleauthority:
Add Paginator Functional Interface and ItemPaginator class for paginating lists of arbitrary data. Has to do with #3.

Y'all mind checking the code out too for style and to ensure it works well with ACF. :)

Here's how it looks when using ACF with it:
```
    @Subcommand("listflares|list|lf")
    @CommandPermission("flaredrops.listflares")
    public void listFlares(CommandSender sender, @Default("1") Integer page) {
        new ItemPaginator<>(
                "&eAvailable Flares",
                Flares.getFlares(),
                (flare) -> flare.getItem().getItemMeta().getDisplayName()
        ).sendPage(sender, page);
    }
```

Looks like so:
[
image
](url)

### DIFF
--- a/src/main/java/co/aikar/commands/ItemPaginator.java
+++ b/src/main/java/co/aikar/commands/ItemPaginator.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2016-2017 Daniel Ennis (Aikar) - MIT License
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining
+ *  a copy of this software and associated documentation files (the
+ *  "Software"), to deal in the Software without restriction, including
+ *  without limitation the rights to use, copy, modify, merge, publish,
+ *  distribute, sublicense, and/or sell copies of the Software, and to
+ *  permit persons to whom the Software is furnished to do so, subject to
+ *  the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be
+ *  included in all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


simpleauthority:
Note to myself for when I have some time to look at this again:

- Remove the interface
- Rename ItemPaginator to Paginator
- Rethink page calculation, think preloading
- Cache a global paginator instance, or no? **Needs decision**
- To async, to sync _and_ async, or to sync only and let user async themselves? **Needs decision**
- Create fluent API for pagination (could be very cool if globally cached paginator is used)

@aikar, and who else should be tagged? :)

simpleauthority:
Going to get some work done on this tonight, probably.

src/main/java/co/aikar/commands/ItemPaginator.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        page = Math.max(0, Math.min(page, maximumPages));
+
+        sender.sendMessage(formatHeader(header, (page + 1), (maximumPages + 1)));
+        for (int i = (PAGE_CAP * page); ((i < (PAGE_CAP * page) + PAGE_CAP) && i < items.size()); i++) {
+            sender.sendMessage(formatLine(stringMapper.apply(items.get(i))));
+        }
+    }
+}
```
kacperduras: omomomomom

kacperduras: omomomom.

simpleauthority:
This is no longer relevant as we don't need a full interface for it. Closing.

PR #43, opened 2017-06-10
Add custom Exception Handlers

MiniDigger:
Add the ability to register exception handlers to handle exceptions that occur while executing a command

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -104,7 +104,13 @@ abstract class CommandManager {
         exceptionHandlers.add(exceptionHandler);
     }
 
-    protected void handleUncaughtException(CommandIssuer sender, List<String> args, Exception e){
-        exceptionHandlers.forEach(handler -> handler.execute(sender, args, e));
+    protected boolean handleUncaughtException(BaseCommand scope, RegisteredCommand registeredCommand, CommandIssuer sender, List<String> args, Throwable t){
+        boolean result = false;
+        for(ExceptionHandler handler : exceptionHandlers){
+            if(handler.execute(scope,registeredCommand,sender,args,t)){
+                result = true;
+            }
+        }
+        return result;
     }
 }
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -25,6 +25,7 @@ package co.aikar.commands;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -92,4 +94,17 @@ public RegisteredCommand createRegisteredCommand(BaseCommand command, String cmd
...
+    /**
+     * Registers an {@link ExceptionHandler} that is called when an exception occurs while executing a command.
+     *
+     * @param exceptionHandler the handler that should handle uncaught exceptions
+     */
+    public void registerExceptionHandler(ExceptionHandler exceptionHandler){
+        exceptionHandlers.add(exceptionHandler);
+    }
+
+    protected void handleUncaughtException(CommandIssuer sender, List<String> args, Exception e){
```
aikar: This signature needs the BaseCommand that triggered the exception, and the RegisteredCommand.

Then the args should be the args for that subcommand.

And maybe Throwable instead of Exception?

core/src/main/java/co/aikar/commands/ExceptionHandler.java
```diff
@@ -0,0 +1,42 @@
...
+public interface ExceptionHandler {
+
+  /**
+   * Called when a exception occurs while executing a command
+   *
+   * @param sender the issuer who send the command
+   * @param args the args he used
+   * @param e the thrown exception
+   */
+  void execute(CommandIssuer sender, List<String> args, Exception e);
```
aikar: See comment on protected void handleUncaughtException

core/src/main/java/co/aikar/commands/RegisteredCommand.java
```diff
@@ -155,6 +155,7 @@ void handleException(CommandIssuer sender, List<String> args, Exception e) {
         } else {
             sender.sendMessage("&cI'm sorry, but there was an error performing this command.");
             this.scope.manager.log(LogLevel.ERROR, "Exception in command: " + command + " " + ACFUtil.join(args), e);
+            this.scope.manager.handleUncaughtException(sender, args, e);
```
aikar: This should be a true/false, and if true,don't send the default message (but we can still always print the log)
So plugins can return false to use the default messaging, or send their own and return true.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -104,7 +104,13 @@ public void registerExceptionHandler(ExceptionHandler exceptionHandler){
         exceptionHandlers.add(exceptionHandler);
     }
 
-    protected void handleUncaughtException(CommandIssuer sender, List<String> args, Exception e){
-        exceptionHandlers.forEach(handler -> handler.execute(sender, args, e));
+    protected boolean handleUncaughtException(BaseCommand scope, RegisteredCommand registeredCommand, CommandIssuer sender, List<String> args, Throwable t){
+        boolean result = false;
+        for(ExceptionHandler handler : exceptionHandlers){
+            if(handler.execute(scope,registeredCommand,sender,args,t)){
```
aikar: i'm not sure we even need more than 1 exception handler to begin with since the manager is bound to the plugins scope.

Let's just get rid of having multiple on the manager, and make it "set default"
Update the manager to be .setDefaultExceptionHandler for method name.

Then, let's also make it so you can do new Command().setExceptionHandler(...);

Then if the command doesn't have its own set, then check if the manager has a default set

PR #59, opened 2017-07-17
Rename Subcommand.java to SubCommand.java

ValidateDev:
It really annoys me that the sub command class is named "Subcommand" instead of "SubCommand".


aikar:
This would be an unnecessary massive API break with lots of work to update implementations.

Plus it's opinionated to spell it with a capital C, with leaning on being incorrect, as it is not 2 words.
https://www.ibm.com/support/knowledgecenter/en/SSLVMB_20.0.0/com.ibm.spss.statistics.help/syn_mixed_emmeans.htm
https://docs.oracle.com/cd/E14004_01/books/SiebCTIAdm/SiebCTIAdm_EventsComms24.html
http://search.cpan.org/~perlancar/Getopt-Long-Subcommand-0.09/lib/Getopt/Long/Subcommand.pm
https://stuff.mit.edu/afs/athena/dept/cron/documentation/Manuals/dell-server-admin/en/DRAC_5/racugaa.htm

It's generally all lowercase, but capital S because it's the first letter.

PR #65, opened 2017-07-24
add german translations

MiniDigger:
This is a draft and up to discussion. Some other german speaker should definitely check those. I tried to find good translations while staying close to the original meaning. 

Should we really translate 'vanished' (I translated it into 'getarnt') or is it more sane it leave it in english since everybody will understand it easier when we use the proper term?

Also, I did not test them ingame or tried to load the file at all since I didn't had time yet to play with that. This is a good test for utf8 support I guess :D

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -47,7 +47,7 @@ public abstract class CommandManager <I, FT, F extends MessageFormatter<FT>> {
     protected Map<String, RootCommand> rootCommands = new HashMap<>();
     protected CommandReplacements replacements = new CommandReplacements(this);
     protected ExceptionHandler defaultExceptionHandler = null;
-    protected Set<Locale> supportedLanguages = Sets.newHashSet(Locale.ENGLISH);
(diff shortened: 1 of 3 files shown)


aikar:
add Locale.whatever here: https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/CommandManager.java#L50

Whichever one is just "de".

languages/minecraft/acf-minecraft_en.properties
```diff
@@ -32,4 +32,3 @@ acf-minecraft.is_not_a_valid_name = Error: <c2>{name}</c2> is not a valid userna
 acf-minecraft.multiple_players_match = Error: Multiple players matched <c2>{search}</c2> <c3>({all})</c3>, please be more specific.
 acf-minecraft.no_player_found_server = No player matching <c2>{search}</c2> is connected to this server.
 acf-minecraft.no_player_found = No player matching <c2>{search}</c2> could be found.
-
```
aikar: Evil person! Files always end with a new line!

MiniDigger: I know, but you had two ;)
github is being weird at rendering it, but there is a CRLF behind the last line

PR #73, opened 2017-10-11
Update ACFExample.java

DTS-MC:
add comments


aikar:
Thanks for the idea. This PR doesn't really help since its more questions than explanations.

I've pushed 8c49eca which adds explanations to nearly everything now.

PR #74, opened 2017-10-14
Add a smexy Hungarian translation

Deathstreams:


aikar:
@Deathstreams Hi, you're missing the acf-core file. That needs to be translated too.

aikar:
@Deathstreams closing. reopen once you provide the other file.

PR #86, opened 2018-01-19
[WIP] Add dependency injection functionality, Closes #85

### LINKED ISSUE #85: Dependency Injection Annotation
Provide a way for the command manager to receive dependencies and provide standard dependency injection abilities

Such as

```java
manager.addDependency(Plugin.class, plugin);
manager.addDependency(MyPlugin.class, plugin);
manager.addDependency(MyObjectManager.class, objManager);

public MyCommandClass extends BaseCommand {
    @Dependency MyPlugin plugin;
    @Dependency MyObjectManager manager;
    // ... commands
}
```

On registration, ACF can fulfill the dependencies as provided, as commonly seen in more featured dependency injection frameworks.

MiniDigger:
also ran some intellij inspections.

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
+++ b/bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
@@ -74,7 +74,7 @@ public class BukkitCommandContexts extends CommandContexts<BukkitCommandExecutio
 
                 throw new InvalidCommandArgument(false);
             }
-            return players.toArray(new OnlinePlayer[players.size()]);
+            return players.toArray(new OnlinePlayer[0]);
         });
         registerIssuerAwareContext(World.class, (c) -> {
             String firstArg = c.getFirstArg();
--- a/bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
+++ b/bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
@@ -74,7 +74,7 @@ public class BukkitCommandContexts extends CommandContexts<BukkitCommandExecutio
 
                 throw new InvalidCommandArgument(false);
             }
-            return players.toArray(new OnlinePlayer[0]);
+            return players.toArray(new OnlinePlayer[players.size()]);
         });
         registerIssuerAwareContext(World.class, (c) -> {
             String firstArg = c.getFirstArg();
--- a/core/src/main/java/co/aikar/commands/CommandContexts.java
+++ b/core/src/main/java/co/aikar/commands/CommandContexts.java
@@ -174,7 +174,7 @@ public class CommandContexts <R extends CommandExecutionContext<?, ? extends Com
                 ACFUtil.sneaky(new IllegalStateException("Weird Command signature... String[] should be last or @Split"));
             }
 
-            String[] result = args.toArray(new String[args.size()]);
+            String[] result = args.toArray(new String[0]);
             args.clear();
             return result;
         });
--- a/core/src/main/java/co/aikar/commands/CommandContexts.java
+++ b/core/src/main/java/co/aikar/commands/CommandContexts.java
@@ -174,7 +174,7 @@ public class CommandContexts <R extends CommandExecutionContext<?, ? extends Com
                 ACFUtil.sneaky(new IllegalStateException("Weird Command signature... String[] should be last or @Split"));
             }
 
-            String[] result = args.toArray(new String[0]);
+            String[] result = args.toArray(new String[args.size()]);
             args.clear();
             return result;
         });
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -390,7 +390,7 @@ public abstract class CommandManager <
      * @throws IllegalStateException when there is already an instance for the provided class registered
      */
     public <T> void registerDependency(Class<? extends T> clazz, T instance){
-        registerNamedDependency(clazz, clazz.getName(), instance);
+        registerDependency(clazz, clazz.getName(), instance);
     }
 
     /**
@@ -403,7 +403,7 @@ public abstract class CommandManager <
      * @param instance the instance of the class that should be injected
      * @throws IllegalStateException when there is already an instance for the provided class registered
      */
-    public <T> void registerNamedDependency(Class<? extends T> clazz, String key, T instance){
+    public <T> void registerDependency(Class<? extends T> clazz, String key, T instance){
         if(dependencies.containsRow(clazz) && dependencies.containsColumn(key)){
             throw new IllegalStateException("There is already an instance of " + clazz.getName() + " with the key " + key + " registered!");
         }
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -23,7 +23,6 @@
 
 package co.aikar.commands;
 
-import co.aikar.commands.annotation.Conditions;
 import co.aikar.locales.MessageKeyProvider;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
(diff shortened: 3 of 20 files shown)


MiniDigger:
possible improvements:
[ ] more default dependencies
[ ] injecting of subtypes, currently only exactly matching types are injected
[ ] allow injection of classes that were not registered via a @Singleton like functionality

bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
```diff
@@ -74,7 +74,7 @@ public BukkitCommandContexts(BukkitCommandManager manager) {
 
                 throw new InvalidCommandArgument(false);
             }
-            return players.toArray(new OnlinePlayer[players.size()]);
```
aikar: Undo this change

core/src/main/java/co/aikar/commands/CommandContexts.java
```diff
@@ -174,7 +174,7 @@
                 ACFUtil.sneaky(new IllegalStateException("Weird Command signature... String[] should be last or @Split"));
             }
 
-            String[] result = args.toArray(new String[args.size()]);
+            String[] result = args.toArray(new String[0]);
```
aikar: Undo this change

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -377,6 +380,71 @@ public void addSupportedLanguage(Locale locale) {
...
+     * Registers an instance of a class to be registered as an injectable dependency.<br>
+     * The command manager will attempt to inject all fields in a command class that are annotated with
+     * {@link co.aikar.commands.annotation.Dependency} with the provided instance.
+     *
+     * @param clazz the class the injector should look for when injecting
+     * @param instance the instance of the class that should be injected
+     * @throws IllegalArgumentException when instance isn't an instance of clazz
+     * @throws IllegalStateException when there is already an instance for the provided class registered
+     */
+    public void registerDependency(Class<?> clazz, Object instance){
```
aikar: Use <T> with Class <T> clazz, T instance, to require it actually satisfies the class its providing for.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -377,6 +380,71 @@ public void addSupportedLanguage(Locale locale) {
...
+     * The command manager will attempt to inject all fields in a command class that are annotated with
+     * {@link co.aikar.commands.annotation.Dependency} with the provided instance.
+     *
+     * @param clazz the class the injector should look for when injecting
+     * @param key the key which needs to be present if that
+     * @param instance the instance of the class that should be injected
+     * @throws IllegalArgumentException when instance isn't an instance of clazz
+     * @throws IllegalStateException when there is already an instance for the provided class registered
+     */
+    public void registerNamedDependency(Class<?> clazz, String key, Object instance){
```
aikar: Same as above for <T>, but just make this registerDependency, since its a different signature

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -377,6 +380,71 @@ public void addSupportedLanguage(Locale locale) {
...
+        if(!clazz.isInstance(instance)){
```
aikar: Not necessary with <T> restrictions

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -377,6 +380,71 @@ public void addSupportedLanguage(Locale locale) {
...
+            throw new IllegalArgumentException(instance.getClass().getName() + " isn't a subclass of " + clazz.getName() + "!");
+        }
+        if(dependencies.containsRow(clazz) && dependencies.containsColumn(key)){
+            throw new IllegalStateException("There is already an instance of " + clazz.getName() + " with the key " + key + " registered!");
```
aikar: Not sure we should throw, but just let it overwrite, maybe return old, like a typical put operation.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -377,6 +380,71 @@ public void addSupportedLanguage(Locale locale) {
...
+        dependencies.put(clazz, key, instance);
+    }
+
+    /**
+     * Attempts to inject instances of classes registered with {@link CommandManager#registerDependency(Class, Object)}
+     * into all fields that are marked with {@link Dependency}.
+     *
+     * @param baseCommand the instance which fields should be filled
+     */
+    public void inject(BaseCommand baseCommand){
```
aikar: make package private

core/src/main/java/co/aikar/commands/RegisteredCommand.java
```diff
@@ -201,7 +201,7 @@ void handleException(CommandIssuer sender, List<String> args, Exception e) {
     @Nullable
     Map<String, Object> resolveContexts(CommandIssuer sender, List<String> args, int argLimit) throws InvalidCommandArgument {
         args = Lists.newArrayList(args);
-        String[] origArgs = args.toArray(new String[args.size()]);
+        String[] origArgs = args.toArray(new String[0]);
```
aikar: Undo change

sponge/src/main/java/co/aikar/commands/SpongeCommandContexts.java
```diff
@@ -119,7 +117,7 @@ public SpongeCommandContexts(final SpongeCommandManager manager) {
 
                 throw new InvalidCommandArgument(false);
             }
-            return players.toArray(new OnlinePlayer[players.size()]);
+            return players.toArray(new OnlinePlayer[0]);
```
aikar: Undo change

core/src/main/java/co/aikar/commands/CommandManager.java
aikar: generic T, github stripped it...

aikar:
```java
public <T> void registerDependency(Class<T> clazz, T instance) {
```

core/src/main/java/co/aikar/commands/CommandManager.java
aikar: I think also more descriptive "injectDependencies"

core/src/main/java/co/aikar/commands/CommandManager.java
MiniDigger: I had this and it didn't work. when I want to pass the plugins class, I can't cast to it, thus throwing a compiler error. https://i.imgur.com/ynHskze.png

MiniDigger:
if some bungee or sponge ppl could suggest default dependencies or the platform that would be great.
suggestions for bukkit are welcome too.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -377,6 +380,66 @@ public void addSupportedLanguage(Locale locale) {
...
+                if(object == null){
+                    throw new UnresolvedDependencyException("Could not find a registered instance of " +
+                            field.getType().getName() + " with key " + key + " for field " + field.getName() +
+                            " in class " + baseCommand.getClass().getName());
+                }
+
+                try {
+                    field.setAccessible(true);
+                    field.set(baseCommand, object);
+                    field.setAccessible(false);
```
aikar: Don't know if theres a reason to set this to false? what if it was already accessible before.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -377,6 +380,66 @@ public void addSupportedLanguage(Locale locale) {
...
+    }
+
+    /**
+     * Attempts to inject instances of classes registered with {@link CommandManager#registerDependency(Class, Object)}
+     * into all fields that are marked with {@link Dependency}.
+     *
+     * @param baseCommand the instance which fields should be filled
+     */
+    void injectDependencies(BaseCommand baseCommand){
+        for (Field field : baseCommand.getClass().getDeclaredFields()) {
```
aikar: If getDeclaredFields is the one that only includes its own fields, then switch to the other. we want to include parent class fields too, but I'm not at a java doc atm.

sponge/src/main/java/co/aikar/commands/SpongeCommandManager.java
```diff
@@ -67,6 +67,10 @@ public SpongeCommandManager(PluginContainer plugin) {
         this.formatters.put(MessageType.INFO, new SpongeMessageFormatter(TextColors.BLUE, TextColors.DARK_GREEN, TextColors.GREEN));
         this.formatters.put(MessageType.HELP, new SpongeMessageFormatter(TextColors.AQUA, TextColors.GREEN, TextColors.YELLOW));
         getLocales(); // auto load locales
+
+        //TODO more default dependencies for sponge
+        registerDependency(plugin.getClass(), plugin);
+        registerDependency(PluginContainer.class, plugin);
```
aikar: Remove the PluginContainer one since per @kashike its not the same thing.

core/src/main/java/co/aikar/commands/CommandManager.java
MiniDigger: when we don't use getDeclaredFields, we can't inject private fields. do we want to just iterate over both?

PR #88, opened 2018-01-23
Added owneronly condition for JDA.

dumptruckman:


jda/src/main/java/co/aikar/commands/JDACommandManager.java
```diff
@@ -47,6 +50,18 @@ public JDACommandManager(JDA jda, CommandConfig defaultConfig, CommandConfigProv
...
         this.logger = Logger.getLogger(this.getClass().getSimpleName());
+
+        if (jda.getAccountType() == AccountType.BOT) {
+            botOwner = jda.asBot().getApplicationInfo().complete().getOwner();
+        } else {
+            botOwner = jda.getSelfUser();
+        }
+
+        getCommandConditions().addCondition("owneronly", context -> {
+            if (context.getIssuer().getEvent().getAuthor() != botOwner) {
```
aikar: Null checks? and isnt there a numerical id we can compare on instead of object equality? safer to avoid object equality.

jda/src/main/java/co/aikar/commands/JDACommandManager.java
```diff
@@ -47,6 +50,18 @@ public JDACommandManager(JDA jda, CommandConfig defaultConfig, CommandConfigProv
...
+                throw new ConditionFailedException("Must be bot owner");
```
aikar: Lets use a bit more descriptive wording "Only the bot owner can use this command", and add a // TODO: MessageKey

jda/src/main/java/co/aikar/commands/JDACommandManager.java
dumptruckman: What is going to be null there?

PR #91, opened 2018-01-24
Improvements for JDA implementation.

dumptruckman:


dumptruckman:
I don't know why the 769765f commit is in there... How do I fix this?

PR #99, opened 2018-02-07
Added support for loading message bundles from other ClassLoaders, an…

TehNeon:
…d made the Bukkit & Bungee locales attempt to load from the ClassLoader that ACF is apart of, and the ClassLoader that the CommandManagers plugin belongs to.

Basically gives support for my pull request in locales (https://github.com/aikar/locales/pull/2), and makes it so Bukkit/Bungee plugins search not only in the direct ClassLoader that ACF is apart of, but also the ClassLoader that the CommandManagers plugin instance is located.

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/BukkitLocales.java
+++ b/bukkit/src/main/java/co/aikar/commands/BukkitLocales.java
@@ -48,6 +48,14 @@ public class BukkitLocales extends Locales {
         addMessageBundles("acf-minecraft", pluginName, pluginName.toLowerCase());
     }
 
+    @Override
+    public void addMessageBundle(String bundleName, Locale locale) {
+        // Load our bundles from the ClassLoader which ACF resides in
+        super.addMessageBundle(bundleName, locale);
+        // Attempt to load our bundles from the ClassLoader which the managers plugin resides in
+        this.addMessageBundle(this.manager.getPlugin().getClass().getClassLoader(), bundleName, locale);
+    }
+
     /**
      * Loads the given file
      * @param file
--- a/bukkit/src/main/java/co/aikar/commands/BukkitLocales.java
+++ b/bukkit/src/main/java/co/aikar/commands/BukkitLocales.java
@@ -39,6 +39,7 @@ public class BukkitLocales extends Locales {
     public BukkitLocales(BukkitCommandManager manager) {
         super(manager);
         this.manager = manager;
+        this.addBundleClassLoader(this.manager.getPlugin().getClass().getClassLoader());
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


core/src/main/java/co/aikar/commands/Locales.java
```diff
@@ -118,9 +118,14 @@ public void addMessageBundles(String... bundleNames) {
     }
 
     public void addMessageBundle(String bundleName, Locale locale) {
-        if (!loadedBundles.containsEntry(bundleName, locale)) {
-            loadedBundles.put(bundleName, locale);
-            this.localeManager.addMessageBundle(bundleName, locale);
+        this.addMessageBundle(this.getClass().getClassLoader(), bundleName, locale);
+    }
+
+    public void addMessageBundle(ClassLoader classLoader, String bundleName, Locale locale) {
```
aikar: Need to re-approach this now, what if same bundle name is in 2 different class loaders?

bukkit/src/main/java/co/aikar/commands/BukkitLocales.java
```diff
@@ -48,6 +48,14 @@ public void loadLanguages() {
         addMessageBundles("acf-minecraft", pluginName, pluginName.toLowerCase());
     }
 
+    @Override
```
aikar: You missed the Sponge and JDA versions, but I also would rather us not do it in this fashion.

The Locales class (in ACF) should maintain a list of class loaders, that's initially just its own, and then implementing classes should add their additional class loader IF its different than ACF's, so we can avoid scanning delays.

TehNeon: I had forgotten Sponge and JDA existed, sorry about that. But yeah, it's a bit iffy having to manually register the ClassLoaders per-component of ACF. Maybe we should have some sort of way to just register/add the ClassLoaders then ACF will also scan those environments only IF it has yet to find a ResourceBundle which matches the name in its own ClassLoader environment.

core/src/main/java/co/aikar/commands/Locales.java
TehNeon: I had thought about this when I wrote it, and had this idea but then it skipped over me as I was just thinking of the direct usage in a Bukkit/Bungee environment where it automatically searches for "acf-pluginname" and such. 

So something that could be potentially done is make loadedBundles a Map with the key being a ClassLoader instead, and having the value as the HashMultimap with the name of the bundle and the locale. Unless if you/anyone else has any other ideas on going around this.

TehNeon:
These two commits should fill in for the requested changes.

Issue #96, opened 2018-01-27
Set Default Command Completion by Context/Class


aikar:
Similar to context, be able to set default command completions for all parameters of a context

```java
manager.getCompletions().setDefaultCompletion("@players", OnlinePlayer.class, Player.class); // Class should be varargs, 
```

When Completion is not set for a parameter on a command, then check the default map to find the default completion.

DarkPaladinUHC:
I am going to attempt to implement this today.

DarkPaladinUHC:
should CommandComletions#setDefaultCompletion() return the CommandCompletionHandler that is set to be default?

DarkPaladinUHC:
If a completion id that is not registered is specified, What type of exception should be thrown? `CommandCompletionTextLookupException`?

aikar:
this was recently finished.

PR #101, opened 2018-02-08
WIP: Set default completion contexts

DarkPaladinUHC:
Closes: #96

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandCompletions.java
+++ b/core/src/main/java/co/aikar/commands/CommandCompletions.java
@@ -39,6 +39,7 @@ import java.util.stream.IntStream;
 public class CommandCompletions <C extends CommandCompletionContext> {
     private final CommandManager manager;
     private Map<String, CommandCompletionHandler> completionMap = new HashMap<>();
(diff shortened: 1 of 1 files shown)


core/src/main/java/co/aikar/commands/CommandCompletions.java
```diff
@@ -71,6 +72,27 @@ public CommandCompletionHandler registerAsyncCompletion(String id, AsyncCommandC
         return this.completionMap.put("@" + id.toLowerCase(), handler);
     }
 
+    public CommandCompletionHandler setDefaultCompletion(String id, Class... classes) {
+        // Since you don't have to specify the @ with the id anywhere else
+        if(!id.startsWith("@")) {
```
aikar: This is not desirable. non @ prefixed completions are valid, they are hard coded.

DarkPaladinUHC: Alright, I have removed that.

aikar:
Per our discussion on Discord, since no further progress has been made on this, let's hold off on doing this work, as I have plans for some major restructuring of core code that might invalidate any work done here.

PR #103, opened 2018-02-18
Add brazilian portuguese translation

Dallabona:
Add brazilian portuguese translation


aikar:
Hello, thank you for the contribution!

For the core project, I am wanting to stick to non country specific dialects for the most part, as if we bundled every sub dialect too, it would result in us shipping a lot of languages increasing jar sizes that people don't need.

Can these translations be suitable as a just portuguese? If so, please rename to just portuguese / pt - and make any adjustments needed.

I am wanting to reserve country dialects for plugin authors to handle themselves.

Dallabona:
Hi aikar,

Brazilian portuguese it's a little bit different of portuguese from Portugal. But i don't think it's a problem. So i renamed to just portuguese.

Thank you.

PR #104, opened 2018-02-28
Add support for per issuer locale in Sponge and BungeeCord

games647:
This adds the missing support for issuer locales in Sponge and BungeeCord.

I'm open for feedback. Here is a short summary.
* Added update task and delayed join listener in BungeeCord
    * Similar to the implementation in Bukkit but without reflection
* Added settings change listener in Sponge
    * No task needed. It fires an event every time the players changes the settings
* Moved setLocale method up to the CommandManager to remove the amount of duplicated code
    * Sponge, Bukkit and BungeeCord would have a similar implementation only the UUID call is different. (See below)
    * The Discord implementation doesn't uses it and will always use the fallback locale.
* Added Optional<UUID> getUniqueId to CommandIssuer. 
    * Do you think `@Nullable` is better?
    * Or should we add another generic in CommandManager for player objects, but what should we do then in Discord which doesn't have player classes.
    * Discord only has `long` ids. So it currently returns `Optional.empty()`

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandIssuer.java
+++ b/core/src/main/java/co/aikar/commands/CommandIssuer.java
@@ -25,8 +25,8 @@ package co.aikar.commands;
 
 import co.aikar.locales.MessageKey;
 import co.aikar.locales.MessageKeyProvider;
+import org.jetbrains.annotations.NotNull;
 
-import java.util.Optional;
 import java.util.UUID;
 
 public interface CommandIssuer {
@@ -54,9 +54,9 @@ public interface CommandIssuer {
     }
 
     /**
-     * @return the UUID of that player or Optional.empty() if it's not a player
+     * @return the unique id of that issuer
      */
-    Optional<UUID> getUniqueId();
+    @NotNull UUID getUniqueId();
 
     /**
      * Has permission node
--- a/core/src/main/java/co/aikar/commands/CommandIssuer.java
+++ b/core/src/main/java/co/aikar/commands/CommandIssuer.java
@@ -26,6 +26,9 @@ package co.aikar.commands;
 import co.aikar.locales.MessageKey;
 import co.aikar.locales.MessageKeyProvider;
 
+import java.util.Optional;
+import java.util.UUID;
+
 public interface CommandIssuer {
     /**
      * Gets the issuer in the platforms native object
@@ -49,6 +52,11 @@ public interface CommandIssuer {
     default void sendMessage(String message) {
         getManager().sendMessage(this, MessageType.INFO, MessageKeys.INFO_MESSAGE, "{message}", message);
     }
+
+    /**
+     * @return the UUID of that player or Optional.empty() if it's not a player
+     */
+    Optional<UUID> getUniqueId();
 
     /**
      * Has permission node
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -27,6 +27,7 @@ import co.aikar.commands.annotation.Dependency;
 import co.aikar.locales.MessageKeyProvider;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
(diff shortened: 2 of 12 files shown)


core/src/main/java/co/aikar/commands/CommandIssuer.java
```diff
@@ -50,6 +53,11 @@ default void sendMessage(String message) {
         getManager().sendMessage(this, MessageType.INFO, MessageKeys.INFO_MESSAGE, "{message}", message);
     }
 
+    /**
+     * @return the UUID of that player or Optional.empty() if it's not a player
+     */
+    Optional<UUID> getUniqueId();
```
aikar: We don't use Optionals in ACF. I'm actually going to say lets make this NotNull also.

For JDA, we can use UUID.fromString(String.valueOf(user.userId));

This way we can get a consistent way to identify a user, to be used by other future things I have planned.

Even if an implementation doesnt have authentication to its commands, they can either use session ID's, ie a UUID per connection to the server (RCON),  or better yet, per ip address.

And worst case, random. But it's easy to get creative on identifying same user.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -344,7 +349,32 @@ public void notifyLocaleChange(I issuer, Locale oldLocale, Locale newLocale) {
         });
     }
 
+    public Locale setPlayerLocale(IT player, Locale locale) {
+        I commandIssuer = getCommandIssuer(player);
+        Optional<UUID> uniqueId = commandIssuer.getUniqueId();
+
+        if (uniqueId.isPresent()) {
```
aikar: don't need check here anymore. api should be not null

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -344,7 +349,32 @@ public void notifyLocaleChange(I issuer, Locale oldLocale, Locale newLocale) {
...
+            Locale old = issuersLocale.put(uniqueId.get(), locale);
+            if (!Objects.equals(old, locale)) {
+                this.notifyLocaleChange(commandIssuer, old, locale);
+            }
+
+            return old;
+        }
+
+        throw new IllegalArgumentException("The argument " + player + " doesn't implement getUniqueId. We can only " +
```
aikar: so we can remove this, and note that im trying to avoid game terminology in -core

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -344,7 +349,32 @@ public void notifyLocaleChange(I issuer, Locale oldLocale, Locale newLocale) {
...
+                "set the locale for player objects");
+    }
+
     public Locale getIssuerLocale(CommandIssuer issuer) {
+        Optional<UUID> uniqueId = issuer.getUniqueId();
+        if (usingPerIssuerLocale() && uniqueId.isPresent()) {
```
aikar: same here, no null check needed.

core/src/main/java/co/aikar/commands/CommandIssuer.java
aikar: additionally, add in some validation on creation of Command Issuers (ie where the .createCommandIssuer method is called), that checks .getUniqueId() for null and throws an error "Command Issuer did not return a UUID".

And this method can default MethodNotImplemented exception

games647: >For JDA, we can use UUID.fromString(String.valueOf(user.userId));

I guess you mean `UUID.nameUUIDFromBytes(user.getId().getBytes(StandardCharsets.UTF_8))`, because `UUID.fromString` requires a 128 bit UUID string like this `0aaa2c13-922a-411b-b655-9b8c08404695`. 

I implemented the first one now for non-players in game platforms (Sponge, Bukkit, BungeeCord) using the sender name (i.e "Rcon" or "Console"). 

For JDA we could use a constant `0L` prefix and use `long user.getIdLong()` for the least significant bits. (`new UUID(0L, userId`)

PR #112, opened 2018-03-23
Use MethodHandles for invocation

Proximyst:
As Java 7 introduced [`MethodHandle`s](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/MethodHandle.html), ACF should also use them to invoke as it is faster than reflection. It finds all the data it needs upon creating it, while reflection does so when it's needed. This sure is a mix and match case for needs, but as all commands are registered to be used, it'd be smarter to use a little more processing upon registration than when it's required.

I've most likely missed something, but everything I found boiled down to the same class and the same method, so I'm assuming that's all it needs.


JanTuck:
Amazing as always @Proximyst

Proximyst:
Seems like when I let this slip into #114, it apparently also pushed any changes here. Luckily both are just positive, so I most likely won't need to do more about it, but notice me if that is wanted.

PR #115, opened 2018-03-23
Re-inject into existing members upon re-registering a dependency

### LINKED ISSUE #93: Live Update Dependencies
Given:
```java
class A implements Foo {}
class B implements Foo {}

class Command extends BaseCommand {
    @Dependency Foo foo;
}
manager.registerDependency(Foo.class, new A());
BaseCommand command = new Command();
manager.registerCommand(command);
// command.foo == instance of A
// 3 minutes later
manager.registerDependency(Foo.class, new B());
// command.foo should be updated to now be the instance of B
```

Anytime dependencies are set, re-apply the injection phase and update the dependency to the latest value.

Proximyst:
Closes #93.

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -448,10 +448,6 @@ public abstract class CommandManager <
      * @throws IllegalStateException when there is already an instance for the provided class registered
      */
     public <T> void registerDependency(Class<? extends T> clazz, String key, T instance){
-//        Commented away due to closing issue #93 (https://github.com/aikar/commands/issues/93)
-//        if(dependencies.containsRow(clazz) && dependencies.containsColumn(key)){
-//            throw new IllegalStateException("There is already an instance of " + clazz.getName() + " with the key " + key + " registered!");
-//        }
         dependencies.put(clazz, key, instance);
 
         if (!rootCommands.isEmpty()) {
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -440,9 +440,10 @@ public abstract class CommandManager <
      * @throws IllegalStateException when there is already an instance for the provided class registered
      */
     public <T> void registerDependency(Class<? extends T> clazz, String key, T instance){
-        if(dependencies.containsRow(clazz) && dependencies.containsColumn(key)){
-            throw new IllegalStateException("There is already an instance of " + clazz.getName() + " with the key " + key + " registered!");
-        }
+//        Commented away due to closing issue #93 (https://github.com/aikar/commands/issues/93)
+//        if(dependencies.containsRow(clazz) && dependencies.containsColumn(key)){
+//            throw new IllegalStateException("There is already an instance of " + clazz.getName() + " with the key " + key + " registered!");
+//        }
 
         dependencies.put(clazz, key, instance);
     }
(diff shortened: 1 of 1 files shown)


MiniDigger:
Why didn't you remove the old code?

Proximyst:
Honestly, don't really know. Made some sense to me while writing that I should just keep it for some reason. I'll remove it fully in a second.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -483,6 +510,20 @@ void injectDependencies(BaseCommand baseCommand) {
...
+    protected <T> void reinjectDependencies(Class<? extends T> type, Class<?> owner, Object instance, T newValue) {
+        Set<MethodHandle> handles = dependencyFields.get(type, owner);
+        if (handles.isEmpty()) {
+            return;
+        }
+        for (MethodHandle handle : handles) {
+            try {
+                handle.invoke(instance, newValue);
+            } catch (Throwable throwable) {
+                throwable.printStackTrace(); //TODO should we print our own exception here to make a more descriptive error?
```
aikar: manager.log("Error reinjecting deps", throwable);

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -473,6 +491,15 @@ void injectDependencies(BaseCommand baseCommand) {
                             field.setAccessible(true);
                         }
                         field.set(baseCommand, object);
+                        {
```
aikar: we got rid of the other method handle stuff as it didnt help, but for setters i guess im ok with it if it works without errors.

however, this approach simply does not work.

Above all this code does comparison for ID/Key.

You need to store the key with the type + handle, and the reinjection needs to only apply on matching keys.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -483,6 +510,20 @@ void injectDependencies(BaseCommand baseCommand) {
         } while (!clazz.equals(BaseCommand.class));
     }
 
+    protected <T> void reinjectDependencies(Class<? extends T> type, Class<?> owner, Object instance, T newValue) {
+        Set<MethodHandle> handles = dependencyFields.get(type, owner);
+        if (handles.isEmpty()) {
+            return;
+        }
+        for (MethodHandle handle : handles) {
+            try {
```
aikar: per above comment, we must only match on keys. It's prob easiest to just use Fields instead of MethodHandles

core/src/main/java/co/aikar/commands/CommandManager.java
aikar: I think id rather us just use the field, so its easier to look up the fields key in the reinjection.

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -68,6 +75,7 @@ public synchronized CommandOperationContext peek() {
         }
     });
     protected Map<String, RootCommand> rootCommands = new HashMap<>();
+    protected Table<Class<?>, Class<?>, Set<MethodHandle>> dependencyFields = HashBasedTable.create();
```
aikar: Going to need to be Table<Class<?>, String, Set<Field>>
Table<Dependency Type, Dependency Key, Set of Fields>

And can we move it down to beside the existing dependencies.

aikar:
Closing due to inactivity. Please create a new PR on latest master if you get a chance to try this again and please see my last feedback.

PR #114, opened 2018-03-23
Fix parsing, more safety, add named MessageChannels and format the code properly

Proximyst:
I seem to have let https://github.com/aikar/commands/pull/112 slip into this, but I guess that's fine, as long as it's accepted. I can revert it on this branch if wanted.
This fixes https://github.com/aikar/commands/issues/113 and makes the contexts a bit more safe when it comes to e.g. banning users, as it no longer can pick one user of e.g. 10 who have the same name (possible, as if I name myself Aikar#0002, it'd be 100% accepted by Discord and all bots).

From now on, all bots will need to manually get the formatted message from the event if they wish to display or log it somewhere the displayname is wanted, as all bots get unformatted messages (i.e. `**<@0138917313> fat _italics_**` instead of `@nickname fat italics`).

PR #116, opened 2018-03-23
[WIP] Documentation, optimisation and formatting

Proximyst:
**This is WIP:** It is only PR-ed so far in order to raise awareness, in case anyone else would like to PR into my fork with documentation which would get a large PR into aikar's repository.

The entire PR is based off of documenting, (mini-)optimising and formatting the code. The goal is obviously to have all methods, classes, fields, and so on documented, though that probably won't happen. Not all annotations are documented either so far, as I don't know what all of them actually do.

The JDA module is left untouched until there is an update to whether #114 gets merged or not.
I'll put in the extra effort to make my other PRs obsolete and have it all in one giant PR if that is wanted, as I don't personally know how well git does merging all code when I've reformatted it all.

The documentation route goes as follows:
 1. Core
 1. Bukkit
 1. Spigot
 1. BungeeCord
 1. PaperSpigot
 1. Sponge
 1. JDA when the update has happened

Any feedback and all reviews are welcome during the work, and any contributions would be greatly appreciated.

Issues which I work off of: https://github.com/Proximyst/commands-1/issues/


paper/src/main/java/co/aikar/commands/PaperAsyncTabCompleteHandler.java
```diff
@@ -70,7 +70,8 @@ public void onAsyncTabComplete(AsyncTabCompleteEvent event) {
             }
             event.getCompletions().addAll(results);
             event.setHandled(true);
-        } catch (Exception ignored) {}
+        } catch (Exception ignored) {
```
MiniDigger: Can't we keep brackets on one line for cases like this?

Proximyst: That should be fixed with https://github.com/aikar/commands/pull/116/commits/919f940a4f39255490e9691717fbd360b8923083 and https://github.com/aikar/commands/pull/116/commits/14a4487337471e91791328877003d909d3be31b3

bukkit/src/main/java/co/aikar/commands/BukkitCommandIssuer.java
```diff
@@ -82,8 +82,8 @@ public boolean hasPermission(String name) {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (this == o) { return true; }
```
Snow-Pyon: This is the case where I hate 1TBS, wouldn't it be better to use K&R for this kind of stuff? In my opinion, this looks a lot better:
```
if (this == o)
    return true;
```
@aikar thoughts on this?

Proximyst: I'll change it around to allow for non-oneliners if either @aikar or another member could provide some more insights on their view on this. I'd still like to comply with Google's standard of using braces, though:
```java
if (this == o) {
    return true;
}
```

MiniDigger: I would use `if (this == o) return true;` over the other solutions proposed here

Proximyst: In that case, I'd like to wait for either aikar or another member to take a go before I change anything in the style guide on the PR.

Proximyst: I've asked for @bramhaag's insights, and he said as follows:
```
[10:53] bramhaag: whatever you do
[10:53] bramhaag: don't do if (this == o) { return true; }
[10:54] bramhaag: 
if (this == o) {
    return true;
}
 for consistency
[10:54] bramhaag: or 
if (this == o)
    return true;
 for aesthetics
```

I guess it boils down to those two at last, so I'd like to simply propose to modify Google's standard to use the second ("aesthetically appealing") version.

aikar: I prefer braces but def dont wanna do 1 line braces,

if () {
 }

please.

This was all generated by IDE is why it ended up like that.

core/src/main/java/co/aikar/commands/ACFPatterns.java
```diff
@@ -49,17 +49,28 @@
...
-
-
-    private ACFPatterns() {}
     @SuppressWarnings("Convert2MethodRef")
     static final Map<String, Pattern> patternCache = ExpiringMap.builder()
-            .maxSize(200)
-            .expiration(1, TimeUnit.HOURS)
-            .expirationPolicy(ExpirationPolicy.ACCESSED)
-            .build();
+                                                                .maxSize(200)
```
aikar: Drop the .builder to next line so we can keep the rest aligned the same, I really hate super far indentations.

core/src/main/java/co/aikar/commands/ACFUtil.java
```diff
@@ -34,18 +34,43 @@
...
+        put(1, "I");
+    }};
 
-    private ACFUtil() {}
+    /**
+     * The constructor is not supposed to be used.
+     * If it is absolutely necessary to use, you'll need to obtain an instance of {@link sun.misc.Unsafe} on Java 8.
+     */
+    private ACFUtil() {
+        throw new SecurityException(); // deny all instances
```
aikar: this isnt necessary

core/src/main/java/co/aikar/commands/ACFUtil.java
```diff
@@ -421,19 +480,13 @@ public static String combine(String[] args, int start) {
...
-            case "t":
-            case "true":
-            case "on":
-            case "y":
-            case "yes":
-            case "1":
-                return true;
-        }
-        return false;
+        test = test.toLowerCase();
```
aikar: revert this. switch statements are faster/better aligned for this check.

core/src/main/java/co/aikar/commands/ACFUtil.java
```diff
@@ -499,41 +554,21 @@ public static boolean isNumber(String str) {
...
     public static String intToRoman(int integer) {
-        if (integer == 1) {
-            return "I";
-        }
-        if (integer == 2) {
-            return "II";
-        }
-        if (integer == 3) {
-            return "III";
+        boolean useParens = integer > 1000; // e.g. (X) = 10*1000, as the parentheses is the "computerised" overbear
```
aikar: this is btter off as a switch instead of using a map

core/src/main/java/co/aikar/commands/Annotations.java
```diff
@@ -49,8 +49,16 @@
...
 
-    String getAnnotationValue(AnnotatedElement element, Class<? extends Annotation> annoClass, int options) {
-        Annotation annotation = element.getAnnotation(annoClass);
+    private static boolean hasOption(int bitmaskOptions, int option) {
+        return (bitmaskOptions & option) == option;
+    }
+
+    @Override
+    String getAnnotationValue(AnnotatedElement object, Annotation annotation, int options) {
+        if (annotation == null) { // allow for using a simple #getAnnotation instead of handling more in the usage
```
aikar: revert this please and the matching change. 

using .getAnnotation() will not be allowed anymore in ACF, everything must route through this code

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -64,77 +67,202 @@
...
+     */
     void onRegister(CommandManager manager, String cmd) {
         manager.injectDependencies(this);
         this.manager = manager;
 
         final Annotations annotations = manager.getAnnotations();
         final Class<? extends BaseCommand> self = this.getClass();
 
-        String[] cmdAliases = annotations.getAnnotationValues(self, CommandAlias.class, Annotations.REPLACEMENTS | Annotations.LOWERCASE | Annotations.NO_EMPTY);
+        String[] cmdAliases = null;
```
aikar: revert this

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -161,17 +289,23 @@ void onRegister(CommandManager manager, String cmd) {
...
+    /**
+     * This recursively registers all subclasses of the command as subcommands, if they are of type {@link BaseCommand}.
+     * Any classes which are not annotated with {@link Ignored} are taken into account.
+     *
+     * @param cmd
+     *         The command name of this command.
+     */
     private void registerSubclasses(String cmd) {
         for (Class<?> clazz : this.getClass().getDeclaredClasses()) {
+            if (clazz.isAnnotationPresent(Ignored.class)) { continue; }
```
aikar: I don't want this. We already have BaseCommand check. This isn't necessary

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -203,6 +341,7 @@ private void registerSubcommands() {
 
         for (Method method : this.getClass().getMethods()) {
             method.setAccessible(true);
+            if (method.isAnnotationPresent(Ignored.class)) { continue; }
```
aikar: again this is not necessary. This can be done in the future with processors if someone wants.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -374,7 +525,6 @@ public void execute(CommandIssuer issuer, String commandLabel, String[] args) {
             if (args.length > 0) {
                 CommandSearch cmd = findSubCommand(args);
                 if (cmd != null) {
-                    execSubcommand = cmd.getCheckSub();
```
aikar: revert this

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -396,39 +546,93 @@ public void execute(CommandIssuer issuer, String commandLabel, String[] args) {
...
+     *
+     * @return The context which is being registered to the {@link CommandManager}'s {@link
+     * CommandManager#commandOperationContext thread local stack}.
+     */
     private CommandOperationContext preCommandOperation(CommandIssuer issuer, String commandLabel, String[] args, boolean isAsync) {
         Stack<CommandOperationContext> contexts = CommandManager.commandOperationContext.get();
         CommandOperationContext context = this.manager.createCommandOperationContext(this, issuer, commandLabel, args, isAsync);
         contexts.push(context);
         lastCommandOperationContext = context;
-        execSubcommand = null;
```
aikar: revert this

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -479,26 +695,62 @@ private void executeCommand(CommandOperationContext commandOperationContext,
...
+     * @param issuer
+     *         The issuer to take into account.
+     * @param cmd
+     *         The command to check whether the issuer can execute.
+     *
+     * @return Whether the issuer can execute the command.
+     */
     public boolean canExecute(CommandIssuer issuer, RegisteredCommand<?> cmd) {
-        return true;
+        return cmd.hasPermission(issuer);
```
aikar: not needed. permission checks were already done. this was old api that will prob go away and meant to be used for misc stuff.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -479,26 +695,62 @@ private void executeCommand(CommandOperationContext commandOperationContext,
...
+     *         The command name as entered by the user instead of the ACF registered name.
+     * @param args
+     *         All arguments entered by the user.
+     * @param isAsync
+     *         Whether this is run off of the main thread.
+     *
+     * @return The possibilities to tab complete in no particular order.
+     */
+    public List<String> tabComplete(CommandIssuer issuer, String commandLabel, String[] args, boolean isAsync) {
+        if (!hasPermission(issuer)) {
```
aikar: not needed, handled at manager

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -515,39 +767,62 @@ public boolean canExecute(CommandIssuer issuer, RegisteredCommand<?> cmd) {
...
         final Set<String> cmds = new HashSet<>();
         final int cmdIndex = Math.max(0, args.length - 1);
         String argString = ApacheCommonsLangUtil.join(args, " ").toLowerCase();
+
         for (Map.Entry<String, RegisteredCommand> entry : subCommands.entries()) {
             final String key = entry.getKey();
             if (key.startsWith(argString) && !CATCHUNKNOWN.equals(key) && !DEFAULT.equals(key)) {
                 final RegisteredCommand value = entry.getValue();
-                if (!value.hasPermission(issuer)) {
+                if (!canExecute(issuer, value)) {
```
aikar: while it seems ok in principle, this is a breaking change as one expects permission checks to already be done for canExecute, and anyone overriding it may of not implemented a permission check (This applies for the above revert request too)

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -694,9 +987,10 @@ String getCheckSub() {
 
         @Override
         public boolean equals(Object o) {
-            if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
+            if (this == o) { return true; }
+            if (!(o instanceof CommandSearch)) { return false; }
```
aikar: revert this. This violates java specs

core/src/main/java/co/aikar/commands/annotation/CommandAliases.java
```diff
@@ -0,0 +1,35 @@
...
+package co.aikar.commands.annotation;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * This is an internal annotation in order to use repeated {@link CommandAlias} annotations.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+public @interface CommandAliases {
```
aikar: dont want this

core/src/main/java/co/aikar/commands/annotation/Ignored.java
```diff
@@ -0,0 +1,34 @@
...
+package co.aikar.commands.annotation;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * This is applied to any subcommands which are to be ignored when registering.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Ignored {
```
aikar: per the comments on use, remove this, unneeded.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -64,69 +65,228 @@
...
-    String commandName;
-    String permission;
-    String conditions;
 
-    private ExceptionHandler exceptionHandler = null;
-    CommandOperationContext lastCommandOperationContext;
-    private String parentSubcommand;
+    /**
+     * Creates a new instance of {@link BaseCommand} without any settings applied.
+     * You will have to manually write to {@link #commandName}.
```
aikar: This isnt accurate. it will generate a command name automatically for you on register

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -465,8 +704,20 @@ private CommandSearch findSubCommand(String[] args, boolean completion) {
         return null;
     }
 
-    private void executeCommand(CommandOperationContext commandOperationContext,
-                                CommandIssuer issuer, String[] args, RegisteredCommand cmd) {
+    /**
+     * Executes the command properly with checks in order.
```
aikar: the wording is just weird here. it just executes the given command.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -329,43 +532,19 @@ private void registerSubcommand(Method method, String subCommand) {
     }
 
     /**
-     * Takes a string like "foo|bar baz|qux" and generates a list of
-     * - foo baz
-     * - foo qux
-     * - bar baz
-     * - bar qux
+     * Executes this command properly by handling it through other methods (look at the see fields).
```
aikar: def need better wording.

This is the primary entry point for integrations to take raw input and map it to a command to execute.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -265,29 +432,57 @@ private void registerSubcommands() {
...
         }
+
         Class<?> clazz = method.getDeclaringClass();
         String parent = getParentSubcommand(clazz);
+
         return parent == null || parent.isEmpty() ? sub : parent + " " + sub;
     }
 
+    /**
+     * Gets the parental subcommand(s) of the input Class.
```
aikar: need better wording. this calculates if the parent command is set to a specific subcommand (ie, all methods share a common subcommand), and if so, combine all the parents to find our current subcommand. so a has inner b, be has inner c.

if a has subcommand a, b has b, then c.getParent() will be 'a b'

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -64,69 +65,228 @@
...
-        return origArgs;
+    private static List<String> filterTabComplete(String arg, List<String> cmds) {
+        return cmds.stream()
+                   .distinct()
+                   .filter(cmd -> cmd != null && (arg.isEmpty() || ApacheCommonsLangUtil.startsWithIgnoreCase(cmd, arg)))
+                   .collect(Collectors.toList());
     }
 
+    /**
+     * This sets the parenting command of this.
```
aikar: i think this method is unused. the wording is super confusing. we need to jus remove this setter as itll never be API nor should it change.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -64,69 +65,228 @@
...
+    CommandManager<?, ?, ?, ?, ?, ?> manager = null;
+
+    /**
+     * The command which owns this. This may be null if there are no owners.
+     */
+    @Nullable BaseCommand parentCommand = null;
+
+    /**
+     * A map of all registered commands under this.
+     * A registered command is any subcommand.
```
aikar: lets drop the subcommand wording.

A registered command is not a subcommand. It is simply a command.

Subcommand is simply one access pattern to reach the command. But you can also use an alias and access it from a root command.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -64,69 +65,228 @@
...
     public static final String CATCHUNKNOWN = "__catchunknown";
+
+    /**
+     * This is a field which contains the key in the {@link #subCommands} map for the method which is default for the
+     * entire (sub)command.
+     */
     public static final String DEFAULT = "__default";
+
+    /**
+     * Stores all the subcommands of this command.
```
aikar: same as below, drop the subcommand wording. its just "Stores all commands registered to this command group."

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -64,69 +65,228 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+/**
+ * BaseCommand stores data about a single command tree.
+ * This includes {@link CommandManager who} manages it, what {@link RegisteredCommand subcommands} it has and so on.
```
aikar: Per comments below. Should be noted that base commands are more so a "Command Group", that shares common elements (subcommand paths, config, intent, permissions, etc)

base command is a command group of registered commands.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -396,18 +575,46 @@ public void execute(CommandIssuer issuer, String commandLabel, String[] args) {
         }
     }
 
+    /**
+     * Gets the registered subcommand of the given arguments.
```
aikar: again here. drop subcommand notion

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -396,18 +575,46 @@ public void execute(CommandIssuer issuer, String commandLabel, String[] args) {
...
+        origArgs = new String[] {};
     }
 
+    /**
+     * This is ran before any command operation has been performed.
+     *
+     * @param issuer
+     *         The user who executed the command.
+     * @param commandLabel
+     *         The label the user used to execute the command. This is not the command name, but their input.
```
aikar: maybe note specifically "when there is multiple aliases, which alias was used"

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -479,26 +730,62 @@ private void executeCommand(CommandOperationContext commandOperationContext,
         }
     }
 
+    /**
+     * Whether the command can be executed by the issuer.
+     * <p>
```
aikar: Actually go ahead and mark this deprecated and @see Conditions. This API was how you could do generic conditions before ,and now we have a proper api for that.

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -531,37 +831,86 @@ public boolean canExecute(CommandIssuer issuer, RegisteredCommand<?> cmd) {
...
-
-    private static List<String> filterTabComplete(String arg, List<String> cmds) {
-        return cmds.stream()
-                   .distinct()
-                   .filter(cmd -> cmd != null && (arg.isEmpty() || ApacheCommonsLangUtil.startsWithIgnoreCase(cmd, arg)))
-                   .collect(Collectors.toList());
+        return filterTabComplete(args[args.length - 1], cmds);
     }
 
+    /**
```
aikar: again lets drop the subcommand wording.

lets rename method "getCommandBySubcommand"

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -531,37 +831,86 @@ public boolean canExecute(CommandIssuer issuer, RegisteredCommand<?> cmd) {
...
+     *
+     * @return The subcommand found or null if none.
+     */
     RegisteredCommand getSubcommand(String subcommand) {
         return getSubcommand(subcommand, false);
     }
 
+    /**
+     * Gets a subcommand under the given name.
+     * If requireOne is true, it won't accept more than a single matching subcommand.
```
aikar: same as above, and rename method

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -531,37 +831,86 @@ public boolean canExecute(CommandIssuer issuer, RegisteredCommand<?> cmd) {
...
+     * @param args
+     *         All arguments given by the issuer.
+     *
+     * @return Whether it found a command or not.
+     *
+     * @see #executeCommand(CommandOperationContext, CommandIssuer, String[], RegisteredCommand)
+     * @see #getSubcommand(String)
+     * @see RegisteredCommand#invoke(CommandIssuer, List, CommandOperationContext)
+     */
     private boolean executeSubcommand(CommandOperationContext commandContext, String subcommand, CommandIssuer issuer, String... args) {
```
aikar: rename method to findAndExecuteCommand

core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -694,9 +1074,14 @@ String getCheckSub() {
 
         @Override
         public boolean equals(Object o) {
-            if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
+            if (this == o) {
+                return true;
+            }
+            if (!(o instanceof CommandSearch)) {
```
aikar: still need to undo this

core/src/main/java/co/aikar/commands/annotation/CatchUnknown.java
```diff
@@ -26,6 +26,13 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Gets all subcommands which are unknown at the time. This means any arguments which are unresolved and have no
```
aikar: wording is just super confusing here.

this is to mark a method that should be executed anytime an unknown command is encountered.

core/src/main/java/co/aikar/commands/annotation/CommandPermission.java
```diff
@@ -26,6 +26,9 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Sets the permission required to perform this command.
```
aikar: Note that permission format will vary by implementation

core/src/main/java/co/aikar/commands/annotation/Default.java
```diff
@@ -26,6 +26,10 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Sets a default value to use. If this is applied to a method, it'll be the default command handler.
```
aikar: "default command handler for the root command of this group"

core/src/main/java/co/aikar/commands/annotation/Flags.java
```diff
@@ -26,6 +26,9 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Sets the flags each parameter uses in this method.
```
aikar: Flags are used to configure and change behavior of context resolution. For adding restrictions on a resolved context, see Conditions

core/src/main/java/co/aikar/commands/annotation/HelpCommand.java
```diff
@@ -26,7 +26,14 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Sets the help command handler. Whenever the help menu is invoked, this will be called.
```
aikar: This caries an implicit @CatchUnknown, @Default and @Subcommand("help|h|?")

aikar: But only if the command group does not define those annotations itself

core/src/main/java/co/aikar/commands/annotation/HelpCommand.java
```diff
@@ -26,7 +26,14 @@
...
+ */
 @Retention(RetentionPolicy.RUNTIME)
 public @interface HelpCommand {
+    /**
+     * The command aliases of the help command.
```
aikar: This is not correct. per above, it overrides the subcommands, not command aliases

core/src/main/java/co/aikar/commands/annotation/Split.java
```diff
@@ -27,6 +27,9 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Joins arguments into a single piece of text with the specified separator.
```
aikar: For array based parameters, defines the regex pattern to split on

core/src/main/java/co/aikar/commands/annotation/Subcommand.java
```diff
@@ -26,6 +26,10 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Defines the member this is attached to a subcommand of the parent.
```
aikar: Defines the subcommand that can be used to execute this command. this is appended onto the root command for the command group, as well as any parent command groups subcommand base.

core/src/main/java/co/aikar/commands/annotation/Syntax.java
```diff
@@ -26,6 +26,11 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+/**
+ * Specifies the syntax to be used when executing this command.
+ * It should not include any descriptions of the arguments nor when some are allowed and when they are not.
```
aikar: Use of this annotation is not necessary. Syntax will be automatically generated for you.
Use this annotation to override automatic syntax

core/src/main/java/co/aikar/commands/contexts/IssuerAwareContextResolver.java
```diff
@@ -26,5 +26,13 @@
 import co.aikar.commands.CommandExecutionContext;
 import co.aikar.commands.CommandIssuer;
 
-
-public interface IssuerAwareContextResolver<T, C extends CommandExecutionContext<?, ? extends CommandIssuer>> extends ContextResolver<T, C> {}
+/**
+ * The very same as {@link ContextResolver} except it now also must know the issuer.
```
aikar: Wrong wording. 
ISsuer aware resolvers may conditionally consume input arguments, and might resolve the context purely based on the issuer of the command rather than command input.

Examples would be "What world are you standing in? What Clan are you currently in? What is your current chat channel?"

core/src/main/java/co/aikar/commands/contexts/OptionalContextResolver.java
```diff
@@ -27,6 +27,12 @@
 import co.aikar.commands.CommandIssuer;
 
 /**
- * Context Resolver that can accept null input
+ * The same as {@link ContextResolver}, however it can accept a null context.
```
aikar: append, if the parameter was marked optional, will still be called with an empty args list

aikar:
I've manually pulled out every bit of documentation that was valid in this PR in commit 0be34242dbf04a10ea09113ff76b2d1870c92cc1, and added some documentation of my own for ones that were super wrong in this PR.

I don't really care for the reordering stuff at this time, as I rather do that myself when I don't have an open major branch.

Please open a SMALL new PR for any more updates.

PR #118, opened 2018-03-24
fix #117

### LINKED ISSUE #117: Detailed help header is broken
see https://i.imgur.com/BMtyK4t.png

MiniDigger:


MiniDigger:
I would also want to revert https://github.com/aikar/commands/commit/db6ff7ed2342e2d4605be73d63e3b46459c30a38
or change this
https://github.com/aikar/commands/blob/db94aa43174bf0ed835881e2e2db5021b00d25d8/core/src/main/java/co/aikar/commands/HelpEntry.java#L42
to get rid of the double / but I don't know the motivation behind that commit.
seems like @aikar had an issue, so either his setup is special or mine

Proximyst:
Does `RegisteredCommand#command` contain a slash already? In that case, it shouldn't be non-stripped in any usual setup..

MiniDigger:
to me it looks like in the lang file with have the prefix (/) and the command, but the command itself already has the prefix:
https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/HelpEntry.java#L41-L43
so help entry shouldn't have the prefix I guess, since we want the prefix in the lang file

Issue #123, opened 2018-04-19
add @Private annotation to hide commands in tabcomplete and help


MiniDigger:
Useful for commands triggered by commandblocks or chat events you don't want to player to use directly.

would add a check here https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/BaseCommand.java#L520
and here
https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/CommandHelp.java#L67

I this is something we want in acf I can throw together a PR when I get home today.

aikar:
Implemented in #124

PR #124, opened 2018-04-19
add a new annotation to hide commands from tabcompletion and help

MiniDigger:
implements #123

seems to work fine

### DIFF
--- a/core/src/main/java/co/aikar/commands/BaseCommand.java
+++ b/core/src/main/java/co/aikar/commands/BaseCommand.java
@@ -31,6 +31,7 @@ import co.aikar.commands.annotation.Conditions;
 import co.aikar.commands.annotation.Default;
 import co.aikar.commands.annotation.HelpCommand;
 import co.aikar.commands.annotation.PreCommand;
+import co.aikar.commands.annotation.Private;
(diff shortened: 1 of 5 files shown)


core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -518,6 +519,10 @@ public boolean canExecute(CommandIssuer issuer, RegisteredCommand<?> cmd) {
                     continue;
                 }
 
+                if(value.method.isAnnotationPresent(Private.class)){
```
aikar: Please look up the annotation in the RegisteredCommand ctor, and store a boolean there, and use the annotations class like everything else does

core/src/main/java/co/aikar/commands/CommandHelp.java
```diff
@@ -65,6 +66,11 @@ public CommandHelp(CommandManager manager, RootCommand rootCommand, CommandIssue
             }
 
             RegisteredCommand regCommand = e.getValue();
+
+            if(regCommand.method.isAnnotationPresent(Private.class)){
```
aikar: same as above, reference a boolean on RegisteredCommand here instead

PR #130, opened 2018-05-15
Exception handler take care about logging

OsipXD:
Why if we have exception handler, ACF still loging all exceptions?
For example, I need to throw IllegalArgumentException with text "Value should be a number, can start with + or -". In this case I use exception handler as point where I can mange all error messages and it's cool. I not need to log all exceptions.

And I can just write following code and it will work fine:
```java
manager.setDefaultExceptionHandler((command, registeredCommand, sender, args, t) -> {
        if (t instanceof IllegalArgumentException || t instanceof UnsupportedOperationException) {
            sender.sendError(MessageKeys.ERROR_PREFIX, "{message}", t.getMessage());
            return false;
        }

        return false;
    });
```

### DIFF
--- a/core/src/main/java/co/aikar/commands/RegisteredCommand.java
+++ b/core/src/main/java/co/aikar/commands/RegisteredCommand.java
@@ -177,8 +177,8 @@ public class RegisteredCommand <CEC extends CommandExecutionContext<CEC, ? exten
             try {
                 if (!this.manager.handleUncaughtException(scope, this, sender, args, e)) {
                     sender.sendMessage(MessageType.ERROR, MessageKeys.ERROR_PERFORMING_COMMAND);
+                    this.manager.log(LogLevel.ERROR, "Exception in command: " + command + " " + ACFUtil.join(args), e);
                 }
-                this.manager.log(LogLevel.ERROR, "Exception in command: " + command + " " + ACFUtil.join(args), e);
             } catch (Exception e2) {
                 this.manager.log(LogLevel.ERROR, "Exception in handleException for command: " + command + " " + ACFUtil.join(args), e);
                 this.manager.log(LogLevel.ERROR, "Exception triggered by exception handler:", e2);


aikar:
You are using the wrong exception. Please use [InvalidCommandArgument](https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/InvalidCommandArgument.java)

This exception is special case for handling command feedback and has options for showing usage or not.

This proposed change would break things for people who do want it to log, but don't want to send the user information because it was manually done already.

OsipXD:
Ok. In this case `IllegalArgumentException` can be replaced by `InvalidCommandArgument` because I throw it from my code. But what about `UnsupportedOperationException` that thrown from library and i can't change it type? Cath it inside and throw new `InvalidCommandArgument`? 
But what if I don't want log any exceptions with ACF? 
For example I want delegate it to my own Handler that will write all exceptions to file but not to console.

aikar:
create your own class that extends InvalidCommandArgument named UnsupportedCommand, and throw that.

Alternatively, i'm open to .setDefaultExceptionHandler taking an optional boolean parameter on whether or not to log exceptions if you wish to PR that.

aikar:
no, I want it on the exception handler API, as if someone turns logging off, they should be implementing their own exception handler to do something with it.

OsipXD:
I've tried to add optional boolean parameter but it looks like bad solution. It shoild be part ot ExceptionHandler but it is Interface. Creating boolean field in CommandManager and BaseCommand looks bad too.
In my opinion would be better to create method like `delegateExceptionsLoggingToACF()` (or other name) that will enable old behavior, approve proposed change and write about this breaking change in changelog. Because it looks like bad design.

aikar:
No, I want it to be as I said, .setDefaultExceptionHandler(handler, boolean logExceptions); where setDefaultExceptionHandler(handler) passes a true to that parameter.

It's not bad design, it's perfectly normal design.

OsipXD:
In this case we also should care about non default exception handler, and add this parameter to `BaseCommand#setExceptionHandler(...)`. And will be logical to add this field to ExceptionHandler class (and make it abstract interface instead of interface).

aikar:
No, if someone doesn't want to log or do anything with exceptions, it SHOULD look stupid because that's a stupid thing to do.

In that case they can do `.setDefaultExceptionHandler((command, regCmd, sender, args, t) -> {}, false);`

This essentially force the implementer into saying "I'm turning off the default logging behavior, and here is my alternative behavior that I want to do to handle the errors instead"

having `manager.shouldLogUncaughtExceptions(boolean)` is intuitively reasonable but lets you be naughty and turn that off without passing your own handler, which results in silent errors

I don't want it to be easy to end up with silent errors.

OsipXD:
Oh. My english isn't enough to explain what i mean. I'll try again.
I see two ways:

#### Way 1. Boolean field
Add optional parameter to `CommandManager#setDefaultExceptionHandler(ExceptionHandler)` and to `BaseCommand#setExceptionHandler(...)`.
In this case where store this boolean field? In `CommandManager` and in `BaseCommand`?

#### Way 2. Add ExceptionHandler for logs
Add second exception handler to `CommandManager`, that will be null by default. If it is null or returns false, exception will be logged by ACF.
In this case what use for exception log handler, existing `ExceptionHandler` or create new interface `ExceptionLogHandler`?

aikar:
Yes store it on the manager. It doesn't matter that its stored separately, as that's back to my point - we won't be providing a way to manipulate that boolean without also setting the handler.
API's don't have to worry about what people hacking with reflection will do.

aikar:
I went ahead and added support for this: https://github.com/aikar/commands/compare/9975c3c24ec9...9aac7d780934

PR #134, opened 2018-06-01
Implement permission resolution

simpleauthority:
This is preliminary.

I need to change the manual string searching for the role-based permission option.

Also, somehow we need to find out if the permission check is for a channel. I'm not sure how we could accomplish that. Ideas?


simpleauthority:
See Picture

PR #135, opened 2018-06-02
Implement CommandPermissionResolver

simpleauthority:
This pull request is an initial attempt at implementing permission for commands for the JDA module.

It includes one added file, **JDACommandPermissionResolver** which implements CommandPermissionResolver.

I created a map to store the Discord permission offsets stored mapped by their kebab-cased name (i.e. "ADMINISTRATOR" becomes "administrator", "MANAGE_SERVER" becomes "manage-server".

The implementation of `hasPermission(JDACommandEvent, String)` is fairly simple:

- Get the issuer's Member object
- Check if it's null; if so, return false.
- Get the permission offset from the aforementioned map
- Check if it's null; if so, return false.
- Return whether or not the member has the permission.

---------------------------------------

This PR allows new commands to use the `@CommandPermission` annotation like so:
```java
@CommandAlias("test|t")
public class Command extends BaseCommand {
  @Subcommand("admin|a")
  @CommandPermission("manage-server")
  public void adminCommand(MessageReceivedEvent event) {
    //...
  }
}
```

### DIFF
--- a/jda/src/main/java/co/aikar/commands/JDACommandPermissionResolver.java
+++ b/jda/src/main/java/co/aikar/commands/JDACommandPermissionResolver.java
@@ -1,0 +1,45 @@
+package co.aikar.commands;
+
+
+import net.dv8tion.jda.core.Permission;
+import net.dv8tion.jda.core.entities.Member;
+import net.dv8tion.jda.core.entities.Role;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class JDACommandPermissionResolver implements CommandPermissionResolver {
+    private Map<String, Integer> discordPermissionOffsets;
+
+    public JDACommandPermissionResolver() {
+        discordPermissionOffsets = new HashMap<>();
+        for (Permission permission : Permission.values()) {
+            discordPermissionOffsets.put(permission.name().toLowerCase().replaceAll("_", "-"), permission.getOffset());
+        }
+    }
+
+    @Override
+    public boolean hasPermission(JDACommandEvent event, String permission) {
+        Member guildMember = event.getIssuer().getMember();
+        if (guildMember == null) {
+            return false;
+        }
+
+        // TODO: regex
+        if (permission.startsWith("role.")) {
+            String perm = permission.split("role.")[1];
+            List<Role> roles = event.getIssuer().getJDA().getRolesByName(perm, true);
+            if (roles.size() == 1) {
+                return guildMember.getRoles().contains(roles.get(0));
+            }
+
+            return false;
+        }
+
+        // TODO: We need to check if the event is for a specific channel
+        int permissionOffset = discordPermissionOffsets.get(permission);
+        Permission discordPermission = Permission.getFromOffset(permissionOffset);
+        return guildMember.hasPermission(discordPermission);
+    }
+}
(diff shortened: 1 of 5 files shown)


jda/src/main/java/co/aikar/commands/JDACommandPermissionResolver.java
```diff
@@ -0,0 +1,36 @@
...
+        discordPermissionOffsets = new HashMap<>();
+        for (Permission permission : Permission.values()) {
+            discordPermissionOffsets.put(permission.name().toLowerCase().replaceAll("_", "-"), permission.getOffset());
+        }
+    }
+
+    @Override
+    public boolean hasPermission(JDACommandEvent event, String permission) {
+        Member guildMember = event.getIssuer().getMember();
+        if (guildMember == null) {
```
aikar: bot owner should prob return true here

aikar: this would ensure permission based commands still work in PM.

simpleauthority: Good idea

jda/src/main/java/co/aikar/commands/JDACommandPermissionResolver.java
```diff
@@ -24,6 +24,10 @@ public boolean hasPermission(JDACommandEvent event, String permission) {
             return false;
         }
 
+        if (guildMember.isOwner()) {
```
simpleauthority: Do we have a bot owner field? I could not find one in the JDA docs. Guild owner is the closest I got. If we want bot owner we need to store it ourselves.

aikar: BotOwner is on the JDACommandManager

aikar: make getBotOwnerId public, dunno why thats private

aikar: Also this change is not valid. 

A command may be restricted the bots owner (for a bot that is shared between guilds), but you dont want the guilds owner to access it.

So remove this and simply update line 24 to return if the event user (not guild member) is the owner.

I'm not sure if that getGuild member is valid in context of a PM either.

simpleauthority: Yeah that's why I was hesitant because that would give guild owners * perms, but in the case of sharded bots that's no bueno.

aikar:
@simpleauthority just add JDACommandManager as a parameter to the interface for a permission resolver instead.

jda/src/main/java/co/aikar/commands/JDAOptions.java
```diff
@@ -6,7 +6,7 @@
 public class JDAOptions {
     CommandConfig defaultConfig = new JDACommandConfig();
     CommandConfigProvider configProvider = null;
-    CommandPermissionResolver permissionResolver = new JDACommandPermissionResolver();
```
aikar: per last comment, we just need to pass the manager to the method parameter, so you can revert this back.

simpleauthority:
@aikar I readded the assignment of the default perm resolver to the options object as it feels wrong in the main class. If you want me to revert, I will. As for adding the manager as a parameter, this has been done and the usage of the method was updated.

PR #136, opened 2018-06-02
Reimplement JDAMessageFormatter for Discord

simpleauthority:
Because Discord does not support coloring messages outside of embeddable fields, it is safe to use a simple message formatter than just returns the message itself.

As I was testing my previous PR, I noticed messages coming through with coloring symbols (e.g. `<c1></c1>`), and each only went up to `c3` maximum. Finally, I figured out these were colors that needed to be handled in each case.

To remove colored messages from Discord, the message formatter is passed three empty strings for each of `c1`, `c2`, and `c3` respectively.

Finally, the default formatter is set inside of `JDACommandManager` but can always be overridden by the user.


simpleauthority:
**Before**
Before

**After**
After

PR #139, opened 2018-06-05
Make Users optional arguments

simpleauthority:
This PR allows `User`s to be optional command arguments similar to how the annotation operates on Bukkit's `Player`.


jda/src/main/java/co/aikar/commands/JDACommandContexts.java
```diff
@@ -66,7 +66,11 @@ public JDACommandContexts(JDACommandManager manager) {
             if (c.hasAnnotation(SelfUser.class)) {
                 return jda.getSelfUser();
             }
-            String arg = c.popFirstArg(); // we pop because we are only issuer aware if we are annotated
+            String arg = c.getFirstArg();
+            if (c.hasAnnotation(Optional.class) && (arg == null || arg.isEmpty())) {
```
aikar: This should be c.isOptional()

simpleauthority: Oh is the annotation processed in core?

simpleauthority: In Bukkit, it's checking against whether the context has the annotation.

aikar: Feel free to update the other refs for me then :)

The goal is, that only the init process cares about actual annotations, which is ran through processors. 

then runtime just checks parsed results from the initialization.

it avoids reflection api's during runtime.

also, the annotation processor may make a custom annotation define something as Optional w/o it using the Optional annotation. so for accuracy, it NEEDS c.isOptional()

simpleauthority: Sounds good, I'll edit soon

aikar:
@simpleauthority did you forget about this

simpleauthority:
@aikar yes, Checking it out now.

simpleauthority:
Accidentally didn't configure muh git so buildtools is a contributor now

simpleauthority:
@aikar dis done?

Issue #127, opened 2018-04-25
Replace Guava Immutable Collections


aikar:
https://docs.oracle.com/javase/1.5.0/docs/api/java/util/Collections.html

Use Collections.EMPTY_LIST/MAP/SET where applicable

Issue #120, opened 2018-03-29
Replace Trivial Guava Usages


aikar:
Any trivial guava usage such as Maps.newHashMap etc need to be removed. I want to drop the use of Guava so non minecraft projects aren't forced to use it.

This does not include Table or any other core data structures used, those will be separate issues so it can be done incrementally.

This issue only covers stuff that has simple standard replacements.

PR #142, opened 2018-06-18
Drop trivial Guava usages 

osipxd:
@aikar 
Please check that I did it right.

Relates to #120, #127


paper/src/main/java/co/aikar/commands/PaperAsyncTabCompleteHandler.java
```diff
@@ -65,9 +63,6 @@ public void onAsyncTabComplete(AsyncTabCompleteEvent event) {
 
             BukkitCommandIssuer issuer = this.manager.getCommandIssuer(event.getSender());
             List<String> results = cmd.tabComplete(issuer, commandLabel, args, true);
-            if (event.getCompletions() instanceof ImmutableList) {
```
aikar: Same issue occurs here, just needs to be the java version check now instead. If it returned immutable, we need to rebuild as mutable.

OsipXD: FIxed

Issue #77, opened 2017-11-18
Brigadier Info


aikar:
Don't know if ill need this yet, will see:
https://gist.github.com/Dinnerbone/7370a2846953eee2d8fc64514fb76de8

aikar:
http://wiki.vg/Command_Data

MiniDigger:
If you don't manage to get to it till the end of next week I will hopefully have time to take a look

MiniDigger:
https://github.com/lucko/commodore

Miclebrick:
#weMustGoDeeper

aikar:
First pass goal: Register subcommands independently to brig
2nd pass goal: Register parameters with types (string, or primitive)
3rd pass goal: Register static Command Completions to brig as static completion values

I want to get stuff merged as goals are complete. Lets not try to do it all in 1 PR please.

aikar:
lets also make it like manager.enableUnstableAPI("brigadier") so that its opt in until we know its all stable.

MiniDigger:
I started playing around with it
https://github.com/aikar/commands/compare/master...MiniDigger:feature/brigadier

PR #153, opened 2018-07-24
Add initial brigadier support (implements #77)

MiniDigger:
~this is by no means ready~, just created a place for open points (will not all be tackled in this PR)

- [x] register commands to brigadier
- [x] register params + primitive types to brigadier
- [x] provide static tab completions
- [x] provide dynamic tab completions
- [ ] more complex getArgumentTypeByClazz (support min, max etc)
- [ ] custom argument types (huge stretch) 
- [x] clean up maven dependencies again
- [x] clean up api
- [x] check if it works with perms
- [x] check command blocks

implements #77

this is affected by https://bugs.mojang.com/browse/MC-165562

### how to use:
you just call `commandManager.enableUnstableAPI("brigadier");` on a paper command manager


MiniDigger:
18:36 <+Choco> But the methods you want to register an argument type is ArgumentRegistry.a(new MinecraftKey("namespace:for_argument_type"), YourArgumentType.class, new ArgumentSerializerVoid(SupplerForYourArgumentType));
18:36 <+Choco> As an example, ArgumentRegistry.a(new MinecraftKey("veinminer:tool_type"), ToolArgumentType.class, new ArgumentSerializerVoid<>(ToolArgumentType::tool));

bukkit/src/main/java/co/aikar/commands/BukkitBrigadierManager.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+public class BukkitBrigadierManager<S> extends ACFBrigadierManager<S> {
+
+    public BukkitBrigadierManager(CommandManager<?, ?, ?, ?, ?, ?> manager, CommandDispatcher<S> dispatcher) {
+        super(manager, dispatcher);
+
+        //TODO custom argument types?
+//        registerArgument(Player.class, new ArgumentType<Player>() {
```
aikar: Def not going to do a manual brigadier mapping to every context. we have to map automatically on every registered context.

MrIvanPlays:
Any activity on that ?

aikar:
Some general notes/desires.

What I was hoping for is for this module to have no bukkit references. purely Brigadier Root/Literal/Argument CommandNode references.

Then, I want ACF to Add respective CommandNode children to the RootCommandNode as defined by ACF's metadata.

Then we would add an event in Paper PlayerBuildBrigadierCommands that exposes the RootCommandNode, and ACF then manipulates it (adding children)

Then, the same event can be added to Sponge or any platform using brigadier and this entire module can be utilized over all of them as a shared core.

As I commented on one line, we shouldnt have any kind of direct object type mapping. Just expose the Argument name and basic primitive types (numbers/strings) as its ArgumentNodes

example/pom.xml
```diff
@@ -64,10 +64,10 @@
...
                     <dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
                     <relocations>
-                        <relocation>
-                            <pattern>co.aikar.commands</pattern>
-                            <shadedPattern>com.exampleplugin.acf</shadedPattern>
-                        </relocation>
+                        <!--<relocation>-->
+                            <!--<pattern>co.aikar.commands</pattern>-->
+                            <!--<shadedPattern>com.exampleplugin.acf</shadedPattern>-->
+                        <!--</relocation>-->
```
MiniDigger, line 70: gotta remove this before merging

MiniDigger:
https://i.imgur.com/oyaKVVL.gifv

MiniDigger:
this depends now on https://github.com/PaperMC/Paper/pull/3193 and is thus Paper only

MiniDigger:
now supports perms https://i.imgur.com/Jbi58QL.gifv

MiniDigger:
I would say this is ready for review no 🎉

Sauilitired:
Imgur
Imgur

```java
@Subcommand("find|where") @CommandPermission("hyperverse.find") @CommandAlias("hvf|hvfind") @CommandCompletion("@players") @Description("{@@command.find}") public void findPlayer(final CommandSender sender, final String player) { /* stub */ }
```
It fails on the command alias, but works fine on the subcommand. I am just completing my own name, so the tab completion works fine in the second example. Both commands execute the same, so it's purely a Brigadier thing.

MiniDigger:
@Sauilitired hope you happy now https://i.imgur.com/hKZgVQg.gifv

MiniDigger:
added support for greedy strings https://i.imgur.com/MV64OyJ.gifv

MiniDigger:
https://i.imgur.com/VtUaYrm.png

MiniDigger:
@aikar brigadier now enables automatically if you call `commandManager.enableUnstableAPI("brigadier");` on a **Paper** command manager. 

I currently dont see how we can make the bukkit command manager enable paper stuff when it detects its being used on paper

darbyjack:
> @aikar brigadier now enables automatically if you call `commandManager.enableUnstableAPI("brigadier");` on a **Paper** command manager.
> 
> I currently dont see how we can make the bukkit command manager enable paper stuff when it detects its being used on paper

So if I understand this correctly, if I'm using Paper Manager (which will fallback to Bukkit if they aren't on Paper) and they ARE on Paper, and on a support version for Brigadier, it will attempt to utilize it? Else it will fallback to Paper?

MiniDigger:
nvm am dum, thanks glare. it does that now

I think that was the last thing I wanted to work on! Soooo

:shipit:

paper/pom.xml
```diff
@@ -64,5 +64,11 @@
             <version>1.15.2-R0.1-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.destroystokyo.paper</groupId>
+            <artifactId>paper-mojangapi</artifactId>
+            <version>1.15.2-R0.1-SNAPSHOT</version>
+            <scope>provided</scope>
+        </dependency>
```
MiniDigger, line 72: duplicated dependency @aikar

NiklasEi:
Nice, thanks for the work!
Is it planned at some point to not only support this for Paper?

JOO200:
It should be possible to expose brigadier stuff to waterfall too.
For other server implementations it depends on the server architecture or if they expose brigadier like paper

MrIvanPlays:
> It should be possible to expose brigadier stuff to waterfall too.
> For other server implementations it depends on the server architecture or if they expose brigadier like paper

There's a discussion about it in bungeecord, here: https://github.com/SpigotMC/BungeeCord/issues/2812

MiniDigger:
if you mean support for spigot: that is up to whatever spigot decides to do. I know that I will not hack my way around the limitations spigot introduces (and I doubt aikar will), but if spigot decides to adapt papers api (or something similar), we could consider switching.

but like, this is yet another reason for server admins to upgrade to paper.

MrIvanPlays:
> custom argument types (huge stretch)

_cough_ mojang set limitation _cough_

aikar:
since this is marked unstable, im going to merge for now, as the changes to fix the issues i saw really dont risk plugins and plugins can control if they opt into this or not.

note that issuer aware context resolvers may have funky results.

PR #156, opened 2018-08-03
Deny permissions if command is issued by a webhook

mikroskeem:
And here is the fix.

### DIFF
--- a/jda/src/main/java/co/aikar/commands/JDACommandPermissionResolver.java
+++ b/jda/src/main/java/co/aikar/commands/JDACommandPermissionResolver.java
@@ -22,6 +22,11 @@ public class JDACommandPermissionResolver implements CommandPermissionResolver {
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


aikar:
Copying from discord:
```
this seems like the wrong solution
we need to make permissions support multiple annotations, for OR conditions
and webhooks should have its own permission node
so that you can say @CommandPermission({"webhook", "foo"));
OR
we need a WebhookPermissionResoler
so that you can do stuff like "I know I'm a bridge, figure out permisssible context from the message"
```

mikroskeem:
I find this PR still valid to get rid of bridge users malicously cluttering bot logs with NPE stacktraces.

However I agree that this could have a better solution at some point.

Commands worked fine for bridge users without `@CommandPermission` annotation for now though (which means command wouldn't pass PermissionResolver)

aikar:
ok im fine with merging it then, but we def need to approach this more.
I'll open an issue with my thoughts

PR #157, opened 2018-08-03
Replace Guava multimaps (and Iterables)

mikroskeem:
~~Only thing which isn't done yet is [Locales.java](https://github.com/aikar/commands/blob/3c979a9ef24dc62e6611832cc97ef25f77afec2d/core/src/main/java/co/aikar/commands/Locales.java), since original code calls to `containsEntry(K, V)` and there's no replacement for it in `co.aikar.util.MapSet`~~ (`containsEntry(K, V)` == `get(K).contains(V)`)

Ready for a review.

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/ACFBukkitUtil.java
+++ b/bukkit/src/main/java/co/aikar/commands/ACFBukkitUtil.java
@@ -23,7 +23,6 @@
 
 package co.aikar.commands;
 
-import com.google.common.collect.Iterables;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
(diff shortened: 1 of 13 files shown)


aikar:
erm, I actually already have this done at https://github.com/aikar/commands/tree/remove-multimap

But the problem is there is bugs, likely at my libraries level.

Can you see what the difference is between this PR and my branch and rebase off my branch and do some testing?

mikroskeem:
Oh I didn't notice the work done at all, my bad.

mikroskeem:
I'll split Iterables replacement into separate PR to make migration from Guava simpler

PR #174, opened 2018-09-20
Add unregister method for JDA

simpleauthority:
This PR is to allow users to unregister commands from the `JDACommandManager`

Currently slightly flawed, needs research. When re-registering it spits out a load of messages like so:
```
SEVERE: [ACF] ACF Error: settings registered subcommand boss role for root command settings - but it is already defined in settings
Sep 20, 2018 3:58:31 AM com.algorithmjunkie.freelance.gachapon.acf.JDACommandManager log
SEVERE: [ACF] 2 subcommands of the same prefix may not be spread over 2 different classes. Ignoring this.
```

Not sure if this is just for me, or if it's for all.

### DIFF
--- a/jda/src/main/java/co/aikar/commands/JDACommandManager.java
+++ b/jda/src/main/java/co/aikar/commands/JDACommandManager.java
@@ -161,6 +161,17 @@ public class JDACommandManager extends CommandManager<
         }
     }
 
+    public void unregisterCommand(BaseCommand command) {
+        for (Map.Entry<String, RootCommand> entry : command.registeredCommands.entrySet()) {
+            String commandName = entry.getKey().toLowerCase();
+            JDARootCommand cmd = (JDARootCommand) entry.getValue();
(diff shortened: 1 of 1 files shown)


jda/src/main/java/co/aikar/commands/JDACommandManager.java
```diff
@@ -161,6 +161,17 @@ public void registerCommand(BaseCommand command) {
         }
     }
 
+    public void unregisterCommand(BaseCommand command) {
```
aikar: See how BukkitCommandManager does unregistering.

A single RootCommand can be used by multiple BaseCommands.
You need to remove all subcommands provided by the base command from the root command, and then only remove the root command when it has no subcommands left.

simpleauthority:
I'm dumb and lazy I'll finish this soon.

simpleauthority:
@aikar This works and does not show any errors.

PR #182, opened 2018-12-23
Added Velocity support

Crypnotic:
This PR adds support for the upcoming proxy Velocity.
https://www.velocitypowered.org/

A few note worthy issues:
 - Velocity does not provide a static method for its ProxyServer instance, so the instance has to be passed in by the plugin with `new VelocityCommandManager(proxy, this)`

- Velocity does not provide a way to get the Plugin logger with the provided Plugin or PluginContainer, so it must be fetched using `Logger logger = LoggerFactory.getLogger(plugin.getClass());` UNTESTED

- Velocity uses an annotation to signify the main class, meaning it can't be passed around like the Bukkit and Bungeecord JavaPlugin and Plugin. An Object has to be passed in which the PluginContainer is extracted from it using the Proxy's PluginManager 
`proxy.getPluginManager().getPlugin(plugin.getClass().getAnnotation(Plugin.class).id()).get();`

Any and all feedback is welcomed and appreciated.

*This implementation is essentially a copy/paste/rename. All code belongs to the original author, except the `ACFVelocityUtil#matchPlayer` which was originally written by md_5 and modified to use Velocity classes.*

### DIFF
--- a/velocity/src/main/java/co/aikar/commands/ACFVelocityUtil.java
+++ b/velocity/src/main/java/co/aikar/commands/ACFVelocityUtil.java
@@ -1,0 +1,90 @@
+package co.aikar.commands;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Sets;
+import com.velocitypowered.api.command.CommandSource;
+import com.velocitypowered.api.proxy.Player;
+import com.velocitypowered.api.proxy.ProxyServer;
+
+import net.kyori.text.TextComponent;
+import net.kyori.text.serializer.ComponentSerializers;
+
+public class ACFVelocityUtil {
+
+    @SuppressWarnings("deprecation")
+    public static TextComponent color(String message) {
+        return ComponentSerializers.LEGACY.deserialize(message);
+    }
+
+    public static Player findPlayerSmart(ProxyServer server, CommandIssuer issuer, String search) {
+        CommandSource requester = issuer.getIssuer();
+        String name = ACFUtil.replace(search, ":confirm", "");
+        if (name.length() < 3) {
+            issuer.sendError(MinecraftMessageKeys.USERNAME_TOO_SHORT);
+            return null;
+        }
+        if (!isValidName(name)) {
+            issuer.sendError(MinecraftMessageKeys.IS_NOT_A_VALID_NAME, "{name}", name);
+            return null;
+        }
+
+        List<Player> matches = new ArrayList<>(matchPlayer(server, name));
+
+        if (matches.size() > 1) {
+            String allMatches = matches.stream().map(Player::getUsername).collect(Collectors.joining(", "));
+            issuer.sendError(MinecraftMessageKeys.MULTIPLE_PLAYERS_MATCH, "{search}", name, "{all}", allMatches);
+            return null;
+        }
+
+        if (matches.isEmpty()) {
+            issuer.sendError(MinecraftMessageKeys.NO_PLAYER_FOUND_SERVER, "{search}", name);
+            return null;
+        }
+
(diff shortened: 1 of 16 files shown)


velocity/src/main/java/co/aikar/commands/VelocityCommandCompletions.java
```diff
@@ -0,0 +1,75 @@
...
+import net.kyori.text.format.TextColor;
+
+public class VelocityCommandCompletions extends CommandCompletions<VelocityCommandCompletionContext> {
+
+    public VelocityCommandCompletions(ProxyServer server, CommandManager manager) {
+        super(manager);
+        registerAsyncCompletion("chatcolors", c -> {
+            Stream<TextColor> colors = Stream.of(TextColor.values());
+            if (c.hasConfig("colorsonly")) {
+                colors = colors.filter(color -> color.ordinal() <= 0xF);
```
kashike: This is useless. `TextColor` only has colours, no decorations - decorations are in `TextDecoration`.

velocity/src/main/java/co/aikar/commands/VelocityCommandContexts.java
```diff
@@ -0,0 +1,85 @@
...
+                throw new InvalidCommandArgument(MessageKeys.NOT_ALLOWED_ON_CONSOLE, false);
+            }
+            return proxiedPlayer;
+        });
+
+        registerContext(TextColor.class, c -> {
+            String first = c.popFirstArg();
+            Stream<TextColor> colors = Stream.of(TextColor.values());
+            if (c.hasFlag("colorsonly")) {
+                colors = colors.filter(color -> color.ordinal() <= 0xF);
```
kashike: This is useless. `TextColor` only has colours, no decorations - decorations are in `TextDecoration`.

velocity/src/main/java/co/aikar/commands/VelocityCommandCompletions.java
kashike: ...and likewise, `TextDecoration` only has decorations, not colours.

velocity/src/main/java/co/aikar/commands/VelocityCommandContexts.java
kashike: ...and likewise, `TextDecoration` only has decorations, not colours.

Crypnotic: I realized that the moment I pushed it xD

velocity/src/main/java/co/aikar/commands/ACFVelocityUtil.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    private static Collection<Player> matchPlayer(ProxyServer server, final String partialName) {
+        Preconditions.checkNotNull(partialName, "partialName");
+
+        Optional<Player> exactMatch = server.getPlayer(partialName);
+        if (exactMatch != null) {
+            return Collections.singleton(exactMatch.get());
+        }
+
+        return Sets.newHashSet(Iterables.filter(server.getAllPlayers(), new Predicate<Player>() {
```
astei: Remove the unneeded usage of Guava. We've got Java 8 here:

```java
return server.getAllPlayers().stream()
  .filter(player -> player.getUsername().toLowerCase(Locale.US).startsWith(partialName.toLowerCase(Locale.US)))
  .collect(Collectors.toList());
```

This could be even more efficient with `String#regionMatches()`, but I'll leave that as an exercise to you.

astei: Alternatively, I could just add `matchPlayer()` to the Velocity API.

Crypnotic: I think other people could find a nice use `matchPlayer()`

js6pak:
Defualt help/usage colors doesn't work
image

I'm temporary fixed it by doing this
image
image

Crypnotic:
@aikar AFAIK, the TextColor issue is fixed. However, the fix feels rather dirty so it would be great if there was another way

McLive:
So this might get merged soon? :D Really looking forward to it.

PR #187, opened 2019-02-11
Replace Jodah's ExpiringMap with Guava's Cache

JarvisCraft:
There is no need to keep Jodah ExpiringMap dependency for the project as it can be painlessly replaced with Guava's Cache while the latest is an already used dependency.
This change is split on three commits.
1. The first should **never** brake anything
2. The second one would break *only* build's which relied on Jodah taken implicitly from `acf-core`
3. The third one would also have zero impact except fot the previous case and external usage of package-private `patternCache` (BTW why is it even `package-private?` :) ) which is very unlikely

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -52,12 +52,6 @@
             <version>1.0-SNAPSHOT</version>
             <scope>compile</scope>
         </dependency>
-        <dependency>
-            <groupId>net.jodah</groupId>
-            <artifactId>expiringmap</artifactId>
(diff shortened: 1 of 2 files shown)


aikar:
Sorry for the wasted effort, but we are working on removing all dependency of guava so it can be dropped.
for future PR's of this type please check with me on Discord to make sure it doesn't conflict with plans.

JarvisCraft:
No problems, it didn't take much time <3

PR #188, opened 2019-02-12
BaseCommand#hasPermission checks the required permissions

TehNeon:
Currently the BaseCommand#hasPermission method doesn't actually check if the user has all the required permissions to execute a command, resulting in the following issue:

If user has permission permission.a they can execute:
/example
/example test
/example sub
/example sub testsub

Which is proper but if the user doesn't have permission.a they can only execute:
/example sub testsub

Example code:
```java
@CommandAlias("example")
@CommandPermission("permission.a")
public class ExampleCommand extends BaseCommand {

    @HelpCommand
    public void help(CommandHelp help) {
        help.showHelp();
    }

    @Subcommand("test")
    public void test(CommandSender sender) {
        sender.sendMessage("has permission to test?");
    }

    @Subcommand("sub")
    public class ExampleBCommand extends BaseCommand {

        @Subcommand("testsub")
        public void testSub(CommandSender sender) {
            sender.sendMessage("has permission to testSub?");
        }

    }

    @Subcommand("othersub")
    @CommandPermission("permission.b")
    public class ExampleCCommand extends BaseCommand {

        @Subcommand("othersub")
        public void otherSub(CommandSender sender) {
            sender.sendMessage("has permission to otherSub?");
        }

    }
}
```

### DIFF
--- a/core/src/main/java/co/aikar/commands/BaseCommand.java
+++ b/core/src/main/java/co/aikar/commands/BaseCommand.java
@@ -908,7 +908,7 @@ public abstract class BaseCommand {
     }
 
     public boolean hasPermission(CommandIssuer issuer) {
-        return getRequiredPermissions().stream().allMatch(permission -> manager.hasPermission(issuer, permission)) && (parentCommand == null || parentCommand.hasPermission(issuer));
+        return getRequiredPermissions().stream().allMatch(permission -> manager.hasPermission(issuer, permission));
     }
 
     public Set<String> getRequiredPermissions() {
--- a/core/src/main/java/co/aikar/commands/BaseCommand.java
+++ b/core/src/main/java/co/aikar/commands/BaseCommand.java
@@ -160,6 +160,11 @@ public abstract class BaseCommand {
     @Nullable
     private String parentSubcommand;
 
+    /**
+     * The permissions of the command.
+     */
+    private final Set<String> permissions = new HashSet<>();
+
     public BaseCommand() {
     }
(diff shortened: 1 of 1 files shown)


core/src/main/java/co/aikar/commands/BaseCommand.java
```diff
@@ -908,19 +927,11 @@ public boolean hasPermission(Object issuer) {
     }
 
     public boolean hasPermission(CommandIssuer issuer) {
-        return permission == null || permission.isEmpty() || (manager.hasPermission(issuer, permission) && (parentCommand == null || parentCommand.hasPermission(issuer)));
+        return getRequiredPermissions().isEmpty() || getRequiredPermissions().stream().allMatch(permission -> manager.hasPermission(issuer, permission));
```
hugmanrique: I think this is premature optimization, calling `.stream()` on an empty Set will guarantee `allMatch` returns `true`>\.

This is known as a [vacuous truth](https://en.m.wikipedia.org/wiki/Vacuous_truth)

aikar: < King of micro optimizations ;)

micro optimizations aren't something you "shouldn't" do, just more of a statement that it's hard to justify the time spent on doing it, or shouldn't do it if it makes the code harder to reason.

In this case, I asked him to do this to avoid the object allocation cost of .stream() and makes it clearer on intent without having to understand allMatch behavior for empty.

PR #189, opened 2019-02-12
Switched getField to getDeclaredField for the locale field in BukkitCommandManager

TehNeon:
This resolves an issue where CraftBukkit has the locale field as private.


xknat:
TY 
~initial reporter

PR #193, opened 2019-02-21
Added Method-Getter in RegisteredCommand class

JOO200:
I want to use own annotations for help implementations (link, tooltips, short description etc.)
I can't access the method directly for the annotations, please add the getter ;)


aikar:
Id rather have a .getAnnotation(AnnoClass.class) method instead. I don't want to expose that level of internal.

what you're ultimately wanting to do lines up better with future goals with custom annotation processors though.

JOO200:
Valid point. Changed it.

PR #197, opened 2019-03-03
Add support for CommandPermission annotation as parameter for permiss…

jdkeke142:
…ions by Optional annotation

In the past we could not handle a permission in relation to the presence of an optional argument, it forced us to manage it in the body of the command, and could be repetitive and boring, note that it also works with the annotation Default


core/src/main/java/co/aikar/commands/RegisteredCommand.java
```diff
@@ -226,8 +226,13 @@ void handleException(CommandIssuer sender, List<String> args, Exception e) {
             if (requiresInput && remainingRequired > 0) {
                 remainingRequired--;
             }
+            String parameterPermission = parameter.getPermission();
             if (args.isEmpty() && !(isLast && type == String[].class)) {
                 if (allowOptional && parameter.getDefaultValue() != null) {
+                    if (parameterPermission != null && !parameterPermission.isEmpty() && !sender.hasPermission(parameterPermission)) {
```
aikar: I don't think we need to check if it has a default? we're ultimately saying we're only giving you the permission to supply your own alternative value if you have the permission, else you're forced to use null/default.

if a param was annotated with a default, this would mandate the permission check on every command.

core/src/main/java/co/aikar/commands/RegisteredCommand.java
```diff
@@ -226,8 +226,13 @@ void handleException(CommandIssuer sender, List<String> args, Exception e) {
...
+                        sender.sendMessage(MessageType.ERROR, MessageKeys.PERMISSION_DENIED);
+                        return null;
+                    }
                     args.add(parameter.getDefaultValue());
                 } else if (allowOptional && parameter.isOptional()) {
                     Object value = parameter.isOptionalResolver() ? resolver.getContext(context) : null;
```
aikar: The permission check should be removed from the ele and moved here, and checked if isOptionalResolver()

basically, before we execute a .getContext call is when to check, but don't check when the value is null without even calling a resolver.

aikar: The permission check should also support the multiple perms logic that BaseCommand and RegisteredCommand do.

We should have a Set<String> for getPermission() instead of single string.

The getter in CommandExecutionContext should also represent that as Set<String> return type.

jdkeke142: > The permission check should be removed from the ele and moved here, and checked if isOptionalResolver()
> 
> basically, before we execute a .getContext call is when to check, but don't check when the value is null without even calling a resolver.

I'm not sure to understand what you want to do by moving the code from the else condition here ? You seem to put it in the condition that checks if the list of arguments is empty, but we want to check the permissions precisely if this list is not empty and so that optional arguments are provided ?

.idea/compiler.xml
```diff
@@ -1,51 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
aikar: Please don't delete this.

core/src/main/java/co/aikar/commands/RegisteredCommand.java
aikar: Permission should only be checked when resolver.getContext() is called.

.idea/compiler.xml
jdkeke142: Strange IDE behaviour, my bad

core/src/main/java/co/aikar/commands/RegisteredCommand.java
jdkeke142: I made the changes, does that sound good to you ?

PR #204, opened 2019-03-24
Add bulgarian language

MrIvanPlays:
This commit adds bulgarian language to ACF


dyanakiev:
The translation is accurate. MrIvanPlays is not the most trustable source but he's done good job here.

PR #209, opened 2019-04-26
Add example for limits condition, and clean up example code a bit.

T3hBrian:
The first commit simply cleans up code using IntelliJ's code reformatting, which makes it a bit more readable.

The second commit adds an example for the "limits" condition that was defined in the main class and also shows the syntax for parameters in conditions (or annotations, I suppose.), which I couldn't find anywhere else.

The third commit was just reverting an accidental import, whoops 😅

### DIFF
--- a/example/src/main/java/co/aikar/acfexample/ACFExample.java
+++ b/example/src/main/java/co/aikar/acfexample/ACFExample.java
@@ -23,10 +23,10 @@
 
 package co.aikar.acfexample;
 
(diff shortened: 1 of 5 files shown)


JanTuck:
import co.aikar.commands.annotation.*; Nono

T3hBrian:
> import co.aikar.commands.annotation.*; Nono

Sorry, I didn't know that wildcard imports were a bad idea. Forgot to pull so it ended up being two commits, but how's it looking now?

PR #211, opened 2019-05-03
Declare dependency versions in parent POM

JarvisCraft:
This PR makes all dependencies be declared in parent POM instead of each submodule which makes it easier to keep them up-to-date for the whole project.
As a related feature, submodules are also declared in `dependencyManagement` and variables are used where possible so less stuff should be replaced manually.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,11 @@
         </plugins>
     </build>
 
+    <dependencyManagement>
+        <dependencies>
+        </dependencies>
+    </dependencyManagement>
+
     <dependencies>
         <dependency>
             <groupId>com.google.guava</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -159,6 +159,61 @@
                 <artifactId>acf-jda</artifactId>
                 <version>${project.version}</version>
             </dependency>
+
+            <dependency>
+                <groupId>co.aikar</groupId>
+                <artifactId>minecraft-timings</artifactId>
+                <version>1.0.4</version>
+                <scope>compile</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.bukkit</groupId>
+                <artifactId>bukkit</artifactId>
+                <version>1.12-R0.1-SNAPSHOT</version>
+                <scope>provided</scope>
+            </dependency>
+            <dependency>
+                <groupId>net.md-5</groupId>
+                <artifactId>bungeecord-api</artifactId>
+                <version>1.12-SNAPSHOT</version>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -126,6 +127,96 @@
         </plugins>
     </build>
 
+    <dependencyManagement>
```
aikar: I don't know what the point of this is but I don't like it. dependencies of sub projects do not belong in parent pom. the parent poms primary purpose is to share common configs. acf-bukkit has no need for information on acf-sponge.

MrIvanPlays:
Declaring all dependencies in the parent pom is not a good idea. That's why we have multiple modules. If @aikar wanted to do it in one module, he'd already do it.

pom.xml
JarvisCraft: This is exactly sharing common configs. What this section does is declaring common dependency properties in patent so that those should not be explicitly specified (unless those values should be overridden) in children. `dependencyManagement` by itself will not add those dependencys to project (for this you've got the very `dependencies` tag in patent) but will simply declare that if the child does not specify (e.g. and most commonly) the version of dependency `bar.group-id:foo-artifact-id` then it should be looked up in parent's dependency management section.

The reason why I've put ACF-dependencies in it too is simply to let them also be used (at current it seems to happen inky for ACF-core and ACF-bukkit) in dependent modules without  a need to declare their versions.

JarvisCraft: As of using variables in `<version>` of module, it is a deprecated feature (Maven will allow it but will complain in logs) which is not even needed as children inherit parent's version.

MiniDigger:
@MrIvanPlays can you elaborate why you think its not a good idea? 
its actually pretty common to do so, to have one super pom to define dependencies (we do so at work all the time, spring boot is an extreme example of that https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/pom.xml )

I for one am in favor of this PR.

hugmanrique:
I'm also in favour of this commit. Popular MC related projects such as ProjectAres also take this approach to make dependency updates easier.

JarvisCraft:
@aikar, we need your opinion considering the upper discussion :)

JarvisCraft:
@aikar, needs rereview

aikar:
Sorry but the MC community is the absolute worse source of best practices in software development.

This breaks the separation of concerns principle. 

The only part that was a maybe was the cross-acf dependencies, and even that goes out the water one day when ACF hits 1.0 and version numbers start to vary per sub project.

Issue #89, opened 2018-01-24
Custom Annotation Aliases


aikar:
Provide an API to allow users to register their own annotations to be aliases to default ACF annotations + Configurations: Flags, Conditions, CommandPermission, Subcommand

Something like 

```java
manager.registerAnnotation(CustomAnnotation.class, Conditions.class, "conditionname:conditionconfig");
manager.registerParameterAnnotation(ParameterType.class, CustomAnnotation.class, Conditions.class, "conditionname:conditionconfig");
manager.registerParameterAnnotation(ParameterType.class, CustomAnnotation.class, Flags.class, "flagconfig=flagvalue");
```

Then using `@CustomAnnotation` will be same as putting 
```java
@Conditions("conditionname:conditionconfig") @Flags("flagconfig=flagvalue")
```
 on a parameter or method

I will be personally working this issue, not looking for PR's on it.

simpleauthority:
Do you have a branch for this?

aikar:
yes and no, work is actively being done in master for it.

aikar:
https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/Annotations.java

aikar:
Note the API is not exposed in any way yet as its constantly changing (when I get time to work on it), just doing most of it in master so I can move/refactor things w/o disrupting PR's, and ensure new additions can follow the new pattern.

kumpelblase2:
Have you considered using meta-annotations (so using annotations on annotations) for this instead? Not to mention, basic support would be trivial to implement (given the current state of the code):

In `Annotations.java#L53` if we change this:
```diff
- Annotation annotation = object.getAnnotation(annoClass);
+ Annotation annotation = getAnnotationRecursive(object, annoClass);
```
and define that function like so:
```java
    private static Annotation getAnnotationRecursive(AnnotatedElement object, Class<? extends Annotation> annoClass) {
        if(object.isAnnotationPresent(annoClass)) {
            return object.getAnnotation(annoClass);
        } else {
            for (Annotation otherAnnotation : object.getDeclaredAnnotations()) {
                final Annotation annotationRecursive = getAnnotationRecursive(otherAnnotation.annotationType(), annoClass);
                if(annotationRecursive != null) {
                    return annotationRecursive;
                }
            }
        }
        return null;
    }
```
it would already find the necessary annotations. So we could implement your given example like this:
```java
@Retention(RetentionPolicy.RUNTIME)
[...]

PR #217, opened 2019-06-11
Support simple meta-annotations

kumpelblase2:
This adds basic support for meta-annotations for commands and
parameters and allows users to create their own custom annotations
combining certain existing annotations and values. For example, a user
 can now define their own `CustomAnnotation` which combines a
condition and a flag:

```java
@Conditions("conditionname:confitionconfig")
@Flags("flagconfig=flagvalue")
public @interface CustomAnnotation {}
```
(Necessary `@Retention` and `@Target` ignore for brevity)
And use it just like they would normally.

This works by recursively going through the annotations, instead of
just looking at the root level. The reason most existing annotations
had to be touched is for them to be allowed on other annotations as
most of them were restricted to method or fields.

Currently there's no limit on nesting because if the user wants to
nest it to obscure levels - so be it. We could decide to limit this at
 some point to prevent users from shooting themselves in the foot if
that's necessary.

`@Dependency` and was specifically ignored as I don't think it makes
sense for that to be supported in this use case.

Relates to: #89

### DIFF
--- a/core/src/main/java/co/aikar/commands/Annotations.java
+++ b/core/src/main/java/co/aikar/commands/Annotations.java
@@ -108,9 +108,11 @@ class Annotations <M extends CommandManager> extends AnnotationLookups {
             return object.getAnnotation(annoClass);
         } else {
             for (Annotation otherAnnotation : object.getDeclaredAnnotations()) {
-                final Annotation foundAnnotation = getAnnotationRecursive(otherAnnotation.annotationType(), annoClass);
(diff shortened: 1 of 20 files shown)


kumpelblase2:
Anything I need to adjust for this to move forward?

aikar:
Sorry for delay on this. I saw title and thought this was likely to conflict with future plans, but I dig this and should be good still with my plans.

PR #252, opened 2020-01-26
Console should be able to execute anything

Proximyst:
Under any circumstance, Console should have access to whichever permission ACF requires to execute some form for command.

If a command is executed as one would expect, the current outcome is that console does _not_ have permission to execute it, while a player might very well be able to.


aikar:
This is up to the platform and the implementing permission resolver.

If a permission system wants to enable console to have all permissions, then it should be defined there.

But Bukkit already does this... I have no issues issueing commands on console that has permission requirements, so this is an issue with your permission manager changing this behavior.

PR #258, opened 2020-04-18
Don't allow infinite recursive annotations

JOO200:
Sorry for the mess with the many git commit messages. The only change is the last commit.

Apperently the latest annotation changes can be used to create an infinite recursion loop.
I guess some simple list can break those loops.

I just used the ArrayList as implementation for that Collection. That might be not the best collection implementation for that.
And maybe there should be a information or warning message for the recursion.


core/src/main/java/co/aikar/commands/Annotations.java
```diff
@@ -50,7 +52,7 @@
     }
 
     String getAnnotationValue(AnnotatedElement object, Class<? extends Annotation> annoClass, int options) {
-        Annotation annotation = getAnnotationRecursive(object, annoClass);
+        Annotation annotation = getAnnotationRecursive(object, annoClass, new ArrayList<>()); // TODO: Better collection impl?
```
portlek, line 55: It can be `Collections.emptyList()`

JOO200: Collections.emptyList is immutable. That doesn't make sense for me.

JohnCorby:
the infinite recursion is especially bad for kotlin users, as every class has a kotlin.Metadata annotation, which will lead to an infinite recursion of kotlin.annotation.Target.

this fix would probably also allow you to get rid of the check for the annotation package name starting with "java." because that's kinda a hacky fix anyway imo

core/src/main/java/co/aikar/commands/Annotations.java
portlek, line 55: This would be better option `new ConcurrentLinkedQueue<>()`. There is no any ordered objects, just need `add` and `contains` methods so you can use `Queue`.

aikar: No, this needs to be a HashSet

aikar:
I think the java one wasn't really a "hacky fix" for anything, but simply to avoid checking ones we know have nothing to do with ACF.

Just change this to a set and looks good to me. Sets have good contains performance and more appropriate here.

JOO200:
Ready to merge, Changed colection to HashSet

JohnCorby:
just one more thing:
do we only care about acf annotations? if thats the case, why not just have a check that the package name contains "co.aikar.commands"? then we wouldnt need any of this checking stuff in the first place, right?
actually, sorry that im uninformed, but why do we even need recursion in the first place? everything seemed to work find without it

JOO200:
> do we only care about acf annotations? if thats the case, why not just have a check that the package name contains "co.aikar.commands"? then we wouldnt need any of this checking stuff in the first place, right?
that doesn't work since acf should be relocated by the project and those project could create their own annotations

JohnCorby:
woah. acf can work with custom annotations?

JOO200:
https://github.com/aikar/commands/commit/667dfec904555cddd4a84fb4ccc3650f99304e9b
It was some shortly addition to acf

PR #261, opened 2020-04-29
Independence from the co.aikar.commands package

Etrayed:
Set a few fields, constructors (etc.) to public to enable full use of ACF from other packages.


aikar:
These are intentional as they are not meant to be public API.
If you are using my package scope, you are doing something unsupported and are at risk of breakage.

Please instead discuss your needs on Discord and we can find if a better way to expose the need can be done instead.

PR #262, opened 2020-04-29
Allow properly nesting of messages

### LINKED ISSUE #173: [Improvement] Modifying the Locale system.
What was the confusing thing is the world came down to a simple lack of documentation and understanding of how something worked. 

When using {@@} in description and message, you CAN'T use ROOT values. You MUST supply a children section for the languages. The improvement would be to allow multi-nested values along with root values in the @@

SirYwell:
- Messages can now be defined everywhere instead of second-level only
- The data types string, double, int, long and boolean are (still) supported

This pull request resolves #173. Other data types were supported before, but that is rarely useful I think, as having a `org.bukkit.Location` in your language file would result in an output like `Location{world=CraftWorld{name=world},x=-287.797836225297,y=70.5,z=-13.083220284478912,pitch=-29.70023,yaw=-90.7704}`.

If it's wanted to keep everything that isn't a ConfigurationSection, I'll go with it too and change that part.

Another improvement I could think of otherwise (or if  the string is null) would be to send a warn log message for invalid keys. What do you think about that?

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/BukkitLocales.java
+++ b/bukkit/src/main/java/co/aikar/commands/BukkitLocales.java
@@ -24,7 +24,6 @@
 package co.aikar.commands;
 
 import co.aikar.locales.MessageKey;
-import org.bukkit.configuration.ConfigurationSection;
 import org.bukkit.configuration.InvalidConfigurationException;
 import org.bukkit.configuration.file.FileConfiguration;
 import org.bukkit.configuration.file.YamlConfiguration;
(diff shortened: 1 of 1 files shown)


aikar (commented):
I'm going to merge but i feel like this could of been done better as get/getobject whatever it's called ,and then doing instanceof String or instanceof Number or instanceof Boolean, so it's a single look up vs the 6 lookups it currently is.

would appreciate a follow up pr to improve this.

Sorry for the big delay on reviewing this.

PR #263, opened 2020-04-30
Add name annotation for parameters

JOO200:
As discussed in the discord:
I added a name annotation to change the display name from a parameter.

Either you can use the name annotation directly:
```java
@Subcommand("test")
public class TestCmd extends BaseCommand {
    @Default
    public void onTest(Player executor,
                       @Name("Text")String word,
                       @Name("Anzahl")int amount) {
        executor.sendMessage("Word: \"" + word + "\" and amount " + amount);
    }
}
```

Or you can modify the name by adding the message key to the locale:
```java
manager.getLocales().addMessageStrings(manager.getLocales().getDefaultLocale(), 
                Collections.singletonMap("acf.parameter.text", "Banane"));
```

With @Name annotation the names will be changed:
grafik

It should not affect backwards compatibility


core/src/main/java/co/aikar/commands/RegisteredCommand.java
```diff
@@ -341,7 +333,21 @@ public String getPrefSubCommand() {
...
     public String getSyntaxText() {
-        return syntaxText;
+        return getSyntaxText(null);
+    }
+
+    public String getSyntaxText(CommandIssuer issuer) {
+        if (syntaxText != null) return syntaxText;
+        StringBuilder syntaxBuilder = new StringBuilder(64);
+        for (CommandParameter<?> parameter : parameters) {
+            if (parameter.getSyntax(null) != null) {
```
aikar, line 343: Why call it for null and then recall it with issuer? can't this work with just 1 call?

languages/core/acf-core_hu.properties
```diff
@@ -20,28 +20,28 @@
 #  OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 #  WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
-acf-core.permission_denied=Nincsen jogod a parancs haszn�lat�hoz.
-acf-core.permission_denied_parameter=Nincsen jogod a parancs haszn�lat�hoz.
-acf-core.error_generic_logged=Hiba t�rt�nt. A hiba r�gz�t�sre ker�lt. A kellemetlens�g�rt eln�z�st k�r�nk.
-acf-core.unknown_command=Ismeretlen parancs. K�rlek haszn�ld a <c2>/help</c2> parancsot.
-acf-core.invalid_syntax=Haszn�lat: <c2>{command}</c2> <c3>{syntax}</c3>
+acf-core.permission_denied=Nincsen jogod a parancs használatához.
```
aikar, line 23: lang files shouldn't be in this PR, maybe needs rebase?

PR #266, opened 2020-05-26
Fix CommandCompletionContext.getResolvedArg

JOO200:
Attemp to fix a bug where commandCompletionContext.getResolvedArg returns null for IssuerOnly and IssuerAware inputs

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandCompletionContext.java
+++ b/core/src/main/java/co/aikar/commands/CommandCompletionContext.java
@@ -102,7 +102,7 @@ public class CommandCompletionContext<I extends CommandIssuer> {
             }
         }
         //noinspection unchecked
-        Map<String, Object> resolved = command.resolveContexts(issuer, args, args.size() - 1);
+        Map<String, Object> resolved = command.resolveContexts(issuer, args, name);
         if (resolved == null || paramIdx > resolved.size()) {
             ACFUtil.sneaky(new CommandCompletionTextLookupException());
         }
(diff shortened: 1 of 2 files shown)


JOO200:
I want to push that pr a little bit @aikar 

That bug is still present and that pr should be the solution. Following steps will create the issue:

1. Create your custom completion handler:

```java
commandManager.getCommandCompletions().registerCompletion("dynamic", c -> {
    Player player = c.getContextValue(Player.class);
    if (player == null) return Collections.singletonList("nullValued");
    else return Collections.singletonList(player.getName() + "_found");
});
```

2. use it in a command

```java
    @Subcommand("dynamicCompletion")
    @CommandCompletion("@dynamic")
    public void onDynamicCompletion(Player player, String test) {
        player.sendMessage("You got the text " + test);
    }
```

My pr should fix this. It will make the lookup with the parameter name which is unique. It will resolve the contexts until the needed parameter with the name is matched.

core/src/main/java/co/aikar/commands/CommandCompletionContext.java
```diff
@@ -101,9 +101,13 @@ public boolean hasConfig(String key) {
                 throw new IllegalStateException("Can not find any parameter that can satisfy " + clazz.getName());
             }
         }
+        return getContextValueByName(clazz, name);
+    }
+
+    public <T> T getContextValueByName(Class<? extends T> clazz, String name) throws InvalidCommandArgument {
```
JOO200, line 107: not happy with that. Imo it should be public because somebody might want to use it but without any checks it's a little bit weird. e.g. what happens when no parameter matches this name?

aikar: I think it's probably fine. there is a check for resolved.containsKey(name) on line 110.

aikar:
so with the new commit, is it resolved now?

JOO200:
Yes, couldn't find any issues and the bug is resolved

PR #272, opened 2020-06-10
Add name annotation for parameters

JOO200:
I added a name annotation to change the display name from a parameter.

Either you can use the name annotation directly:
```java
@Subcommand("test")
public class TestCmd extends BaseCommand {
    @Default
    public void onTest(Player executor,
                       @Name("Text")String word,
                       @Name("Anzahl")int amount) {
        executor.sendMessage("Word: \"" + word + "\" and amount " + amount);
    }
}
```

Or you can modify the name by adding the message key to the locale:
```java
manager.getLocales().addMessageStrings(manager.getLocales().getDefaultLocale(), 
                Collections.singletonMap("acf.parameter.text", "Banane"));
```

With @Name annotation the names will be changed:
grafik

It should not affect backwards compatibility

I failed some git-branch things in https://github.com/aikar/commands/pull/263
Now I'm unable to reopen the old pull request, shame on me.


JOO200:
I rebased the pull request on the latest changes from acf.

Currently there is no support for custom names for Brigadier. Currently the command map is created at startup. There is no way to change the messages based on langauges (or I don't see it)

PR #277, opened 2020-07-08
Add support for Kotlin coroutines

kacgal:
Handles exceptions thrown by Kotlin `suspend` functions.


aikar:
I'm not interested in adding kotlin to the build. There has to be a less invasive way to pull this off....

Id suggest trying to achieve this by registering your own context, and maybe setting exception handler.

kacgal:
Do you mean that you don't want kotlin in every final jar of acf, or don't want kotlin at build time at all?
Because this method doesn't include kotlin in the jar, it's completely optional. Just required at build time

knightzmc:
Surely this will cause a failure if Kotlin isn't on the classpath - i.e. if the plugin isn't using it?

kacgal:
No, there is no failure. The JVM does throw exceptions that the classes can't be found, but those are caught and silently ignored - meaning no Kotlin is required at runtime

Nopock:
Lil late but any updates or future thoughts on this?

PR #280, opened 2020-07-23
added @Regex annotation

Machine-Maker:
Added an `@Regex` annotation that can be applied to method parameters. The use cases for this, are small validation requirements, that don't warrant overriding a whole default context for, or registering a condition. Got this idea from Glare asking what the best way to only accept a 2 digit number was.


aikar (changes requested):
I'm not really a fan of this because it can be done as a condition where you register a condition as regex for name then do

`@Condition("regex:Foo[0-9]")` instead.

This is simply another form of conditions but trying to lazily avoid having to register it yourself.

However, future plans would support the idea of registering a custom `@Regex` annotation that maps to `@Condition("regex:xx")` for you and id rather it just be done that way.

Because this is purely convenience over functionality, I'm going to say no to this as i feel this is best suited to user land so that the condition is checked at same place as conditions.

Now, what we can do is simply default register a condition for regex, so everyone can use `@Condition("regex:XXX")` on their parameters out of the box.

So please update the PR to do that.

PR #292, opened 2020-09-22
Fix wrong BungeeCord command suggestions

JOO200:
https://github.com/SpigotMC/BungeeCord/commit/a47b8033857973588a0da98d3ebdc5d970176fa9

This commit added some magic stuff to BungeeCord to check permissions for commands.
ACF already overrides the getPermission method but hasPermission method is untouched.

I had to increase the BungeeCord-Api version to 1.13 but that should be no problem since BungeeCord is backwards compatible to 1.8.

### DIFF
--- a/bungee/pom.xml
+++ b/bungee/pom.xml
@@ -32,7 +32,7 @@
         <dependency>
             <groupId>net.md-5</groupId>
             <artifactId>bungeecord-api</artifactId>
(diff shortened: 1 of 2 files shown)


JOO200:
Changed the way to check permissions to RootCommand$hasAnyPermission

aikar (approved):

PR #298, opened 2020-10-15
Feature/command selectors

DarkEyeDragon:
Adds command selector handling to `BukkitCommandContexts#getOnlinePlayer` However i'm not sure how to deal with @a or selectors that return more than one entity. So in this draft i just return the first entry. However this isn't an optimal solution, and also currently fails if there are no players


bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
```diff
@@ -241,6 +241,20 @@ OnlinePlayer getOnlinePlayer(BukkitCommandIssuer issuer, String lookup, boolean
         Player player = ACFBukkitUtil.findPlayerSmart(issuer, lookup);
         //noinspection Duplicates
         if (player == null) {
+            if (issuer.getManager() instanceof BukkitCommandManager) {
+                BukkitCommandManager manager = (BukkitCommandManager) issuer.getManager();
+                //Check if mc version is higher than 1.13.1 (added in 1.13.2)
+                if ((manager.mcMinorVersion > 13) || (manager.mcMinorVersion == 13 && manager.mcPatchVersion > 1)) {
+                    try {
+                        Entity entity = Bukkit.selectEntities(issuer.getIssuer(), lookup).get(0);
```
aikar, line 249: rather than blind look up 0 relying on errors, iterate and return on first player match instead, and lack of finding should fall through to the lines below.

And we can only run this selectEntities if the input starts with @

bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
```diff
@@ -241,6 +241,20 @@ OnlinePlayer getOnlinePlayer(BukkitCommandIssuer issuer, String lookup, boolean
         Player player = ACFBukkitUtil.findPlayerSmart(issuer, lookup);
         //noinspection Duplicates
         if (player == null) {
+            if (issuer.getManager() instanceof BukkitCommandManager) {
```
aikar, line 244: this is already a bukkit context, this isn't really needed. 

this.manager should already exists.

aikar (changes requested):

bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
DarkEyeDragon: this.manager returns a `CommandManager` which does not have the version fields

PR #302, opened 2020-11-15
Add option to use bot mention as prefix

SuSeu1414:
I added an option to use bot mention as a command prefix. Might be useful when you have per-server prefixes and forget what prefix is used in a particular server. You can enable it with this code:
```java
JDACommandConfig config = new JDACommandConfig();
config.useMentionPrefix(true);
```

It's disabled by default, so it won't change any behaviour until you turn it on

### DIFF
--- a/jda/src/main/java/co/aikar/commands/JDACommandManager.java
+++ b/jda/src/main/java/co/aikar/commands/JDACommandManager.java
@@ -5,6 +5,7 @@ import net.dv8tion.jda.api.AccountType;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.ChannelType;
 import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.SelfUser;
 import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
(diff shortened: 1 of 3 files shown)


jda/src/main/java/co/aikar/commands/JDACommandManager.java
```diff
@@ -247,11 +249,19 @@ void dispatchEvent(MessageReceivedEvent event) {
...
+                prefixFound = "<@!" + selfUserId + ">";
+            }
+        }
         if (prefixFound == null) {
             return;
         }
 
-        String[] args = ACFPatterns.SPACE.split(msg.substring(prefixFound.length()), -1);
+        // str.replaceAll("^[ \t]+", "") - works like String.stripLeading() from Java 11
+        String[] args = ACFPatterns.SPACE.split(msg.substring(prefixFound.length()).replaceAll("^[ \\t]+", ""), -1);
```
MiniDigger, line 264: I would say that this pattern should be cached

SuSeu1414: Ok, now I'm using ```ACFPatterns.getPattern()``` which caches patterns

PR #303, opened 2020-12-18
[WIP] Migrate to Gradle

Prof-Bloodstone:
Initial PR migrating ACF to gradle. It compiles - but would need to be checked further. All feedback welcome!

Things to do:

1. Configure shading as per original root project pom
2. Configure dependency-reduced pom
3. Check if I didn't miss anything (mostly automatic migration)
4. Delete all pom.xml files
5. Cleanup .gitignore?


DavidTheExplorer:
I'm not a maintainer, but why does the library need to switch to gradle?

chickeneer:
> I'm not a maintainer, but why does the library need to switch to gradle?

Primary reason is that certain additional modules would be more easily compatible with a gradle project as opposed to maven. ACF as it is now, would not have an immediate benefit from said migration.

DavidTheExplorer:
> Primary reason is that certain additional modules would be more easily compatible with a gradle project as opposed to maven.
Can you elaborate? Creating maven modules is effortless, and I'm not sure what's the meaning of "easily compatible with a gradle project" - Which project and how?

DavidTheExplorer:
> Primary reason is that certain additional modules would be more easily compatible with a gradle project as opposed to maven. ACF as it is now, would not have an immediate benefit from said migration.

Can you elaborate? Creating maven modules is effortless, and I'm not sure what's the meaning of "easily compatible with a gradle project" - Which project and how?

PR #305, opened 2020-12-31
Add Adventure/MiniMessage as new message parser to acf

JOO200:
I added Adventure as a module to acf. It's still a work in progress.

TODOs:

- [x] Make a proper implementation for MessageType and MessageFormatters. (take a look at the picture below)
- [ ] Check every existing message key so there's no problem with weird color codes.

grafik

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/BukkitCommandManager.java
+++ b/bukkit/src/main/java/co/aikar/commands/BukkitCommandManager.java
@@ -115,7 +115,7 @@ public class BukkitCommandManager extends CommandManager<
         }
 
         try {
-            Class.forName("co.aikar.commands.adventure.text.Component");
+            Class.forName("co.aikar.commands.adventure.adventure.text.Component");
             adventureAvailable = true;
         } catch (ClassNotFoundException ignored) {
             // Ignored
@@ -417,11 +417,13 @@ public class BukkitCommandManager extends CommandManager<
 
     @Override
     public void enableUnstableAPI(String api) {
+        super.enableUnstableAPI(api);
+
         if ("adventure".equals(api) && adventureAvailable) {
+            log(LogLevel.INFO, "Enabling Adventure");
             adventureManager = new ACFBukkitAdventureManager(plugin, this);
             super.adventureManager = adventureManager;
         }
-        super.enableUnstableAPI(api);
     }
 
     public @Nullable ACFBukkitAdventureManager getAdventureManager() {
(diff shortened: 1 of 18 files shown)


JOO200:
I fixed issues with the MessageType and Formatter. I tested the implementation on bukkit and bungee and it works.

I'm not happy with the workaround for the formatter colors but I don't think there's another better solution without loosing backwarts compatibility.

To use the MiniMessage parser:

- Create a BukkitCommandManager or BungeeCommandManager
- enable the unstable api "adventure"
- have fun with MiniMessage.

I modified the example. e.g.
```java
        issuer.sendMessage("<red>This is a <yellow>" +
                "<hover:show_text:'<yellow>Click here'>" +
                "<click:open_url:'https://docs.adventure.kyori.net/minimessage.html'>clickable message</click></hover></yellow>.");
```
will send a message with formatted text.

I exposed the ACFBukkitAdventureManager and ACFBungeeAdventureManager classes (searching better class names) through the BukkitCommandManager and BungeeCommandManager to wrap CommandIssuers but I'm not sure if that's necessary.
You can simply call "XAudiences.create(plugin)" to receive the same (cached) Audience wrapper as they are cached by adventure.

bukkit/src/main/java/co/aikar/commands/BukkitCommandManager.java
```diff
@@ -403,4 +414,19 @@ protected boolean handleUncaughtException(BaseCommand scope, RegisteredCommand r
         }
         return super.handleUncaughtException(scope, registeredCommand, sender, args, t);
     }
+
+    @Override
+    public void enableUnstableAPI(String api) {
+        super.enableUnstableAPI(api);
+
+        if ("adventure".equals(api) && adventureAvailable) {
+            log(LogLevel.INFO, "Enabling Adventure");
```
MiniDigger, line 423: not sure a message here is necessary, ppl already complain about the current messages...

MiniDigger:
might be worse to expose the minimessage instance, so that you can provide your own one, for handling placeholders, custom transformations, whatever.

JOO200:
MiniMessage instance is exposed and I added a context resolver for the Audience class for the command issuer.

I change the draft to a PR now as it's working for me.

GeorgeV220:
Any updates on this?

JOO200:
Short answer: No, there is no update here.
The used MiniMessage version is outdated and it should be updated to a never version.

I use ACF for my own server so I only need to support my platforms. That's why I created my own fork of ACF [here](https://github.com/JOO200/commands/tree/terraconia). I replaced the existing messaging stuff completly with adventure and minimessage and made many changes for my own server. It should compile without any issues but it only supports velocity and paper.

PR #312, opened 2021-01-27
Added option to disable unregistered replacement warning

stijnb1234:
You can now use `manager.hideUnregisteredReplacements(true);` to disable the warning.
The default value is false, so that it will only hide if the user chooses to hide them.

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -75,6 +75,7 @@ public abstract class CommandManager<
     protected CommandHelpFormatter helpFormatter = new CommandHelpFormatter(this);
 
     protected boolean usePerIssuerLocale = false;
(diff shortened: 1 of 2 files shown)


MiniDigger:
whats the usecase for this?

stijnb1234:
> whats the usecase for this?

It is useful in such situations as the following:
- You use a percent sign in your message, for example: "Number: 10%"
- You use PlaceholderAPI variables in your messages, which are replaced by PAPI, and not ACF.
- You have a different system for replacing messages.

Saves a lot of spam in your console.

stijnb1234:
For private plugins this can be an option to temporarily fix spam: https://www.spigotmc.org/resources/console-spam-fix.18410/
However, this is not working for public plugins (you can't force users to use a plugin like this to fix spam).

This PR temporary fixes any spam issues for all plugins, until a better system for placeholders is implemented.

stijnb1234:
Closed because of #313

PR #313, opened 2021-01-30
Adds a new pattern for CommandReplacements: %{<key>}

RoinujNosde:
- Adds a new pattern for CommandReplacements: ```%{<key>}```
- Maintains the old pattern ```%<key>```
- Disables warnings for when the old pattern is not replaced

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandReplacements.java
+++ b/core/src/main/java/co/aikar/commands/CommandReplacements.java
@@ -39,6 +39,7 @@ public class CommandReplacements {
 
     private final CommandManager manager;
     private final Map<String, Map.Entry<Pattern, String>> replacements = new LinkedHashMap<>();
+    private final Map<String, Map.Entry<Pattern, String>> oldReplacements = new LinkedHashMap<>();
 
     CommandReplacements(CommandManager manager) {
         this.manager = manager;
@@ -50,7 +51,7 @@ public class CommandReplacements {
             throw new IllegalArgumentException("Must pass a number of arguments divisible by 2.");
         }
         for (int i = 0; i < replacements.length; i += 2) {
-            addReplacement(replacements[i], replacements[i+1]);
+            addReplacement(replacements[i], replacements[i + 1]);
         }
     }
 
@@ -61,8 +62,10 @@ public class CommandReplacements {
     @Nullable
     private String addReplacement0(String key, String val) {
         key = ACFPatterns.PERCENTAGE.matcher(key.toLowerCase(Locale.ENGLISH)).replaceAll("");
-        Pattern pattern = Pattern.compile("%" + Pattern.quote(key) + "\\b", Pattern.CASE_INSENSITIVE);
+        Pattern oldPattern = Pattern.compile("%" + Pattern.quote(key) + "\\b", Pattern.CASE_INSENSITIVE);
+        Pattern pattern = Pattern.compile("%\\{" + Pattern.quote(key) + "}", Pattern.CASE_INSENSITIVE);
 
+        oldReplacements.put(key, new AbstractMap.SimpleImmutableEntry<>(oldPattern, val));
         Map.Entry<Pattern, String> entry = new AbstractMap.SimpleImmutableEntry<>(pattern, val);
         Map.Entry<Pattern, String> replaced = replacements.put(key, entry);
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/co/aikar/commands/CommandReplacements.java
```diff
@@ -61,7 +61,8 @@ public String addReplacement(String key, String val) {
     @Nullable
     private String addReplacement0(String key, String val) {
         key = ACFPatterns.PERCENTAGE.matcher(key.toLowerCase(Locale.ENGLISH)).replaceAll("");
-        Pattern pattern = Pattern.compile("%" + Pattern.quote(key) + "\\b", Pattern.CASE_INSENSITIVE);
+        Pattern pattern = Pattern.compile("%\\{" + Pattern.quote(key) + "}|%" + Pattern.compile(key) + "\\b",
```
aikar, line 64: the 2nd should be Pattern.quote, not Pattern.compile, surprised this didn't error.

and can we move this to ACFPatterns to use the cache method

core/src/main/java/co/aikar/commands/CommandReplacements.java
```diff
@@ -83,7 +84,7 @@ public String replace(String text) {
         }
 
         // check for unregistered replacements
-        Pattern pattern = Pattern.compile("%.[^\\s]*");
+        Pattern pattern = Pattern.compile("%\\{.[^\\s]*}");
```
aikar, line 87: lets move this to an acfpatterns static const (don't need to use the cache method here, just store result as const in that file and reference it.

though ideally need to move this find unregistered to be done at registration time instead of every single replacement :/

aikar (changes requested):

core/src/main/java/co/aikar/commands/CommandReplacements.java
RoinujNosde: > though ideally need to move this find unregistered to be done at registration time instead of every single replacement :/

Is this possible? At registration time we don't have access to the strings to be replaced.

aikar: Well if instead of using ACFPatterns::getPattern, we store a replacementpattern cache map, and use .computeIfAbsent(), then only run the missing logic in the provider of the computeIfAbsent, then itll only run on the first time its asked for when the pattern is compiled and cached.

core/src/main/java/co/aikar/commands/CommandReplacements.java
RoinujNosde: > and can we move this to ACFPatterns to use the cache method

```java
//addReplacement
String pattern = "%\\{" + Pattern.quote(key) + "}|%" + Pattern.quote(key) + "\\b";

//replace
for (Map.Entry<String, String> entry : replacements.values()) { //changed replacements Map key to String, key is the pattern not compiled
    text = ACFPatterns.getPattern(entry.getKey()).matcher(text).replaceAll(entry.getValue());
}
```

You mean something like this?

aikar: see comment on the other line

RoinujNosde: Hm, already moved then.
The REPLACEMENT_PATTERN to check for unregistered replacements.

stijnb1234 (approved):
Tested it myself, and it works fine. Also fixes the replacement errors for the old pattern. :)

stijnb1234:
@RoinujNosde 

Can you look into this?
"Well if instead of using ACFPatterns::getPattern, we store a replacementpattern cache map in a local field on this class, and use .computeIfAbsent(), then only run the missing logic in the provider of the computeIfAbsent, then itll only run on the first time its asked for when the pattern is compiled and cached."

core/src/main/java/co/aikar/commands/CommandReplacements.java
RoinujNosde: So a <Pattern, String> map, where string is the value to be replaced.
When replacing, we loop through this map, matching the patterns and using the values.

And after that we seek for non-replaced placeholders in the String, if found, compile a Pattern for it and call computeIfAbsent using this pattern as key.
Inside the compute resolver, the unregistered replacements message is logged and the raw pattern is added as value.

@aikar is that what you have in mind?

aikar (approved):
eh I was a bit confused on something thinking this was all one method. 

what I had in mind isn't valid, so going to take this as is and can improve later.

core/src/main/java/co/aikar/commands/CommandReplacements.java
aikar: see final comment

stijnb1234:
@aikar When will this fix be available on the Maven repo?

benwoo1110:
It already merged, so I don't see why it's not available on the latest snapshot version.

PR #318, opened 2021-03-05
Handle numbers in Hexadecimal (0x#) and Binary (0b#) Notation

willies952002:


core/src/main/java/co/aikar/commands/ACFUtil.java
```diff
@@ -466,11 +466,17 @@ public static boolean isTruthy(String test) {
 
 
     public static Number parseNumber(String num, boolean suffixes) {
-        ApplyModifierToNumber applyModifierToNumber = new ApplyModifierToNumber(num, suffixes).invoke();
-        num = applyModifierToNumber.getNum();
-        double mod = applyModifierToNumber.getMod();
+        if (num.matches("^0x([0-9A-Fa-f]*)$")) {
```
stijnb1234, line 469: Use regex pattern caching with the ACFPatterns class here.

https://github.com/aikar/commands/blob/f033dd9541eee468cb05a96db0097ea493f9ec5b/core/src/main/java/co/aikar/commands/ACFPatterns.java#L75

core/src/main/java/co/aikar/commands/ACFUtil.java
```diff
@@ -466,11 +466,17 @@ public static boolean isTruthy(String test) {
...
+            return Long.parseLong(num.substring(2), 16);
+        } else if (num.matches("^0b([01]*)$")) {
```
stijnb1234, line 471: Use regex pattern caching with the ACFPatterns class here.

https://github.com/aikar/commands/blob/f033dd9541eee468cb05a96db0097ea493f9ec5b/core/src/main/java/co/aikar/commands/ACFPatterns.java#L75

stijnb1234 (changes requested):

stijnb1234 (approved):

PR #319, opened 2021-03-24
Add Chinese(Simplified) language support.

Gk0Wk:
A potential problem: Because `LocaleManager.getMessage()` only uses `Locale.getLanguage()` and not `Locale.getCountry()`, `Locale.SIMPLIFIED_CHINESE` and `Locale. TRADITIONAL_CHINESE` will be treated as the same dialect of Chinese. This will cause only one language file to exist for Simplified Chinese and Traditional Chinese.


portlek (approved):

chickeneer:
Merged manually (required rebasing).
Could have messed it up due to also merging in https://github.com/aikar/commands/pull/333

PR #320, opened 2021-04-22
Add constructor based dependency injection

0x4da:
Added constructor dependency injection to the small DI framework that already exists. This is done by adding `CommandManager#registerCommand(Class<? extends BaseCommand> commandClass)` which constructs the class then passes it into the command manager's registerCommand. If the class only has one constructor, it will attempt to use that, if not, it looks for a constructor annotated with @Dependency. Parameters can be annotated with @Dependency to specify the key

### DIFF
--- a/core/src/main/java/co/aikar/commands/CommandManager.java
+++ b/core/src/main/java/co/aikar/commands/CommandManager.java
@@ -28,9 +28,11 @@ import co.aikar.locales.MessageKeyProvider;
 import co.aikar.util.Table;
 import org.jetbrains.annotations.NotNull;
 
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -555,6 +557,66 @@ void injectDependencies(BaseCommand baseCommand) {
...
+
+        Parameter[] parameters = constructor.getParameters();
+        Object[] args = new Object[parameters.length];
+        boolean accessible = constructor.isAccessible();
+        if (!accessible) {
+            constructor.setAccessible(true);
+        }
+        for (int i = 0; i < parameters.length; i++) {
+            String dependency = annotations.getAnnotationValue(parameters[i], Dependency.class);
+            String key = (key = dependency) == null || key.isEmpty() ? parameters[i].getType().getName() : key;
```
MiniDigger, line 600: I feel like this is much more understandable this way
```suggestion
            String key = dependency == null || dependency.isEmpty() ? parameters[i].getType().getName() : dependency;
```

core/src/main/java/co/aikar/commands/CommandManager.java
```diff
@@ -555,6 +557,69 @@ void injectDependencies(BaseCommand baseCommand) {
...
+            if (annotations.hasAnnotation(possibleConstructor, Dependency.class)) {
+                if (constructor != null) {
+                    throw new IllegalStateException("Command " + commandClass.getName() + " has more than one " +
+                            "constructor annotated @Depenedency.");
+                }
+                constructor = possibleConstructor;
+            }
+        }
+        if (commandClass.getConstructors().length == 1) {
+            constructor = commandClass.getConstructors()[0];
```
chickeneer, line 589: This appears really cheap. But strikes me as odd that this takes place after searching for annotated constructors.
Would it make more sense to have this first, with the Annotation check as an else to this? Again, so cheap that it doesn't really matter just looks strange.

Everything else looks good to me though - from just looking at it. Appears eligible for merging soon.

PR #325, opened 2021-05-12
Use every subcommand for command completions

JOO200:
ACF does only use one matching command to complete commands. Therefore overloaded commands doesn't get correctly completed.

This should fix this problem. There is no need to match a specific command here.

filterTabComplete(..) will remove all duplicated completions so there should be no duplicates in the completions.

### DIFF
--- a/core/src/main/java/co/aikar/commands/BaseCommand.java
+++ b/core/src/main/java/co/aikar/commands/BaseCommand.java
@@ -637,9 +637,8 @@ public abstract class BaseCommand {
 
             final List<String> cmds = new ArrayList<>();
             if (search != null) {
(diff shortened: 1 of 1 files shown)


chickeneer:
Sorry for the delay on looking at this PR. 
So I think the changes makes sense and are logical, as the matchCommand should be used during execution to hone in on a single command to be executed. Where completions should consider every possible command.

My only concern, would be that this solution may includes commands for completing which is no longer possible due to number of arguments already input. But that is just from the code analysis level - I probably would have to do some plugin testing to verify one way or the other...

If anyone else has some thoughts on that, it would be helpful. Otherwise I don't feel comfortable merging without some testing on my part.

Issue #335, opened 2021-07-06
Support Velocity 3.0


MAXOUXAX:
[Velocity 3.0 has updated its dependencies and text3 has been completely removed, making Adventure the new dependency to use. So there are a few changes, such as changing the names of some methods and the way components are created.](https://velocitypowered.com/wiki/developers/porting-from-velocity-1/)

These modifications break `acf-velocity@0.5.0-SNAPSHOT` (see error message below)

![image](https://user-images.githubusercontent.com/24844231/124653660-dafa4a80-de9d-11eb-9e32-5f35c48632dc.png)

[A migration guide is available here](https://docs.adventure.kyori.net/migration/text-3.x.html)

MAXOUXAX:
Closing this issue since https://github.com/aikar/commands/pull/336 has been merged

PR #336, opened 2021-07-07
Add velocity 3.0.0 support

AlexProgrammerDE:
Hi! I've added velocity 3.0.0 support with this PR. Here are two changes I made that may need to be looked into:
1. `NamedTextColor.NAMES.values().stream().map(namedTextColor -> namedTextColor)`
The `.map(namedTextColor -> namedTextColor)` was required, so intellij stopped complaining about not assignable types. I am not sure if that is intentional, but it seems to work.
2. 
```java
    @Override
    public CompletableFuture<List<String>> suggestAsync(Invocation invocation) {
        return CompletableFuture.completedFuture(getTabCompletions(manager.getCommandIssuer(invocation.source()), getCommandName(), invocation.arguments()));
    }
```

I am not sure if that needs to be implemented, but I did it either way. 

Let me know if I have to do any more changes to get this PR merged. Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,7 @@
                 </configuration>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>2.22.2</version>
             </plugin>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -32,7 +32,7 @@
 		<dependency>
 			<groupId>com.velocitypowered</groupId>
 			<artifactId>velocity-api</artifactId>
-			<version>1.0.0-SNAPSHOT</version>
+			<version>3.0.0</version>
 			<scope>provided</scope>
 		</dependency>
 	</dependencies>
--- a/velocity/src/main/java/co/aikar/commands/ACFVelocityUtil.java
+++ b/velocity/src/main/java/co/aikar/commands/ACFVelocityUtil.java
@@ -11,14 +11,13 @@ import com.velocitypowered.api.command.CommandSource;
 import com.velocitypowered.api.proxy.Player;
 import com.velocitypowered.api.proxy.ProxyServer;
 
-import net.kyori.text.TextComponent;
-import net.kyori.text.serializer.legacy.LegacyComponentSerializer;
+import net.kyori.adventure.text.TextComponent;
+import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
 
 public class ACFVelocityUtil {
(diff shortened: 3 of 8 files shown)


MAXOUXAX:
This fixes #335 btw :)

sagansfault:
I think `legacyAmpersand()` might be more common in use than `legacySection()`
`TextColor match = NamedTextColor.NAMES.value(ACFUtil.simplifyString(first));` The stream is over `TextFormat`, a super of `TextColor`, matching only to `NamedTextColor` will exclude possible `TextDecoration`
`colors = colors.filter(color -> filters.contains(ACFUtil.simplifyString(color.toString())));` (in completions and other places) `toString()` will use the child implementation but `TextFormat` has more children than `NamedTextColor` and `TextDecoration` which might return weird strings like object hashes

DolphinGamerYT:
I just had issues running acf on velocity, and came here to create a PR fixing them, but just found that they have been fixed. This can be merged perfectly.
Good job @AlexProgrammerDE 😄

AlexProgrammerDE:
Sorry for the late response. Imma share my opinion on @sagansfault's thought.

>     * I think `legacyAmpersand()` might be more common in use than `legacySection()`

I agree that sometimes that is more reasonable, but I try to achieve velocity support and not break any current functionality or add any in that sense. May be worth another PR once this one is merged.

>     * `TextColor match = NamedTextColor.NAMES.value(ACFUtil.simplifyString(first));` 
>The stream is over `TextFormat`, a super of `TextColor`, matching only to `NamedTextColor` will _exclude_ possible `TextDecoration`

TextDecoration isn't a TextColor tho. It is a TextFormat. (A interface TextColor implements too) If you look closely, there is another check below the TextColor code that also concats all the TextDecorations to the stream:
https://github.com/AlexProgrammerDE/commands/blob/master/velocity/src/main/java/co/aikar/commands/VelocityCommandCompletions.java#L47-L49

>     * `colors = colors.filter(color -> filters.contains(ACFUtil.simplifyString(color.toString())));` 
> (in completions and other places) `toString()` will use the child implementation but `TextFormat` has more children than `NamedTextColor` and `TextDecoration` which might return weird strings like object hashes

It does have other children? I don't think we inject any form of other children of TextFormat into any methods. Additionally, the #toString() method should always be overwritten in such kind of implementations. Also, there is no way that we by accident execute the Object#toString() instead of the NamedTextColor#toString() or TextFormat#toString(). Look here for reference: [copy paste heaven](https://stackoverflow.com/questions/15668032/how-to-call-the-overridden-method-of-a-superclass)

Hope that explains everything. Cheers!

PR #340, opened 2021-08-12
update to velocity 3.0.0

sagansfault:
This required a fair decent bit of change, mostly in the command completions and contexts sections.
Previously, `TextFormat` was used primarily as the `@chatcolor` command completion. As of velocity `3.x` using kyori adventure `4.8.x`, the hierarchy has changed. Chat formatting as been split into `NamedTextColor` (a class) for actual colors while text decorations have been placed into `TextDecoration` (an enum), both of which implement the **now empty** interface `TextFormat`. Because this interface is empty, we have no way of common methods for getting names which leads to the addition in a util class:
```java
public static String getTextFormatName(TextFormat textFormat) {
      if (textFormat instanceof NamedTextColor) {
          return ((NamedTextColor) textFormat).toString();
      } else if (textFormat instanceof TextDecoration) {
          return ((TextDecoration) textFormat).toString();
      } else {
          return "";
      }
  }
```
Unfortunately, kyori has made this pretty difficult by splitting these parts up so the above is required. Which leads to something like:
```java
@Subcommand("foo|bar")
@CommandCompletion("@chatcolors")
public void foobar(CommandSource player, TextFormat color) {
    if (color instanceof TextColor) {
        player.sendMessage(Component.text("kyori 👎", ((TextColor) color)));
    } else if (color instanceof TextDecoration) {
        player.sendMessage(Component.text("kyori 👎").decorate(((TextDecoration) color)));
    }
}
```
Comment your thoughts!

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -7,15 +7,15 @@
         <sourceOutputDir name="target/generated-sources/annotations" />
         <sourceTestOutputDir name="target/generated-test-sources/test-annotations" />
         <outputRelativeToContentRoot value="true" />
+        <module name="acf-jda" />
+        <module name="acf-paper" />
+        <module name="example-plugin" />
         <module name="acf-sponge" />
         <module name="acf" />
-        <module name="acf-example" />
-        <module name="example-plugin" />
-        <module name="acf-brigadier" />
-        <module name="acf-paper" />
-        <module name="acf-jda" />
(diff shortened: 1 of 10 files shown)


wafflecoffee:
How is this different from #336 ?

sagansfault:
It's unfortunately not too different. I made the mistake of not checking for an existing PR before making one 😛

PR #348, opened 2021-10-21
Command Visibilty Conditions

PoQuatre:
Hi! I added this feature for one of my plugins, so I thought why not make a PR for it because people might find useful.

As the title suggest it adds a way to programmatically show or hide a subcommand, it only change the tab completion like `@Private`.

### How it works
1. You register a condition :
```java
commandManager.getCommandVisibilityConditions().addCondition("creative", (context) -> {
    if (!context.getIssuer().isPlayer()) return true;
    return context.getIssuer().getPlayer().getGameMode() == GameMode.CREATIVE;
});

commandManager.getCommandVisibilityConditions().addCondition("flying", (context) -> {
    if (!context.getIssuer().isPlayer()) return true;
    return context.getIssuer().getPlayer().isFlying();
});
```
2. You use it with either `@ShowCommand` or `@HideCommand` :
```java
@Subcommand("example1")
@ShowCommand("creative")
public void firstExample(CommandSender sender) {
    ...
}

@Subcommand("example2")
@HideCommand("creative")
public void secondExample(CommandSender sender) {
    ...
}

@Subcommand("example3")
@ShowCommand("creative")
@HideCommand("flying")
public void thirdExample(CommandSender sender) {
    ...
}
```
3. That's all you have to do to make it work.

So what is it doing in game :
- In the first example, the subcommand `example1` is only shown in the tab completion if the player is in gamemode creative, this is because with `@ShowCommand` the subcommand is hidden by default and is only shown when all conditions return true.
- In the second example, the subcommand `example2` is only shown in the tab completion if the player is **not** in gamemode creative, this is because with `@HideCommand` the subcommand is shown by default and is only hidden when all conditions return true.
- And in the third example, the subcommand `example3` is only shown in the tab completion if the player is in gamemode creative **and** not flying, this is because `@HideCommand` takes precedence over `@ShowCommand`.

### Condition Configs
Like normal conditions, configs are available here is an example :
```java
commandManager.getCommandVisibilityConditions().addCondition("height", (context) -> {
    if (!context.getIssuer().isPlayer()) return true;

    final int min = context.getConfigValue("min", -1);
    final int max = context.getConfigValue("max", -1);
    final double height = context.getIssuer().getPlayer().getHeight();

    return min < height && (max == -1 || max > height);
});
```
```java
@Subcommand("example")
@ShowCommand("height:min=64,max=128")
public void example(CommandSender sender) {
    ...
}
```
In this example the subcommand `example` is only shown when the player is between Y 64 and Y 128.

**Note :** You can chain conditions using `|` between them.

### DIFF
--- a/core/src/main/java/co/aikar/commands/BaseCommand.java
+++ b/core/src/main/java/co/aikar/commands/BaseCommand.java
@@ -144,6 +144,14 @@ public abstract class BaseCommand {
      */
     @Nullable String conditions;
     /**
+     * The show conditions of the command. This may be null if no show conditions has been provided.
+     */
+    @Nullable String showConditions;
+    /**
+     * The hide conditions of the command. This may be null if no hide conditions has been provided.
+     */
+    @Nullable String hideConditions;
+    /**
      * Identifies if the command has an explicit help command annotated with {@link HelpCommand}
      */
     boolean hasHelpCommand;
--- a/core/src/main/java/co/aikar/commands/BaseCommand.java
+++ b/core/src/main/java/co/aikar/commands/BaseCommand.java
@@ -32,7 +32,9 @@ import co.aikar.commands.annotation.Conditions;
 import co.aikar.commands.annotation.Default;
 import co.aikar.commands.annotation.Description;
 import co.aikar.commands.annotation.HelpCommand;
+import co.aikar.commands.annotation.HideConditions;
 import co.aikar.commands.annotation.PreCommand;
+import co.aikar.commands.annotation.ShowConditions;
 import co.aikar.commands.annotation.Subcommand;
 import co.aikar.commands.annotation.UnknownHandler;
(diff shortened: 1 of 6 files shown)


chickeneer (commented):
Left a couple of code analysis issues. But big picture. I don't see any real advantages to the Hide condition, based on your description and code analysis it appears redundant in terms of functionality.
Since all Show conditions are required and only one hide condition can prevent seeing. They are effectively the same just different names. Only advantage being the API level of how creating the commands feels which has value. But the flip to that - is that it introduces questions about how stuff interacts together. 

Thinking about it further even. I would kinda wish it was integrated more directly with the Private annotation. As it is is for the same type of behavior. "Private unless this condition is met to show".

 Open to anyone elses thoughts on the matter - I don't see us merging this as is. I definitely see the advantage of having conditions for visibility though.

DevJoey:
Nice feature would be very useful to have more control over that!

Issue #330, opened 2021-06-15
[1.18] Command completions not loading after base command with Brigadier enabled


DarkEyeDragon:
After switching to acf-core instead of acf-paper command completions no longer work after the basecommand

The first is using the acf-paper module
![68747470733a2f2f692e696d6775722e636f6d2f37387058674b702e706e67](https://user-images.githubusercontent.com/6131099/122071314-ef937780-cdf6-11eb-95a2-9ac50d63058f.png)
the second runs acf-core (notice the lack of worlds and users)
![image](https://user-images.githubusercontent.com/6131099/122071334-f4582b80-cdf6-11eb-8afe-6bc7e5e40317.png)
after some further testing it seems like it does autocomplete in console
![image](https://user-images.githubusercontent.com/6131099/122073900-1357bd00-cdf9-11eb-9cda-fb5476c242fb.png)

```java
@Default
@CommandPermission("rtp.teleport.self")
@CommandCompletion("@worlds")
@Description("Teleport players to a random location.")
@Syntax("[world/player] [world]")
public void onTeleport(CommandIssuer sender, @Optional PlayerWorldContext target, @Optional @CommandPermission("rtp.teleport.world") RandomWorld world) {
           //impl
}
```

DarkEyeDragon:
![image](https://user-images.githubusercontent.com/6131099/145493620-5a6b4606-a57c-4c36-90bc-a6a5cccb717c.png)
Providing some more info since the issue still persists

It seems to work fine when using the plugin prefix, but not when using the command directly

DarkEyeDragon:
Well after a long search i figured out it is related to the `CommandManager#enableUnstableAPI("brigadier");` 
not setting this line solves the issue.

PR #357, opened 2022-01-10
Fix @Default method brigadier tab completion

DeathRealms:
Fixes tab completion for the default method when brigadier api is enabled (#330)

### DIFF
--- a/brigadier/src/main/java/co.aikar.commands/ACFBrigadierManager.java
+++ b/brigadier/src/main/java/co.aikar.commands/ACFBrigadierManager.java
@@ -89,10 +89,9 @@ public class ACFBrigadierManager<S> {
         root = rootBuilder.build();
         boolean isForwardingCommand = rootCommand.getDefCommand() instanceof ForwardingCommand;
 
-        if (rootCommand.getDefaultRegisteredCommand() != null) {
-            CommandNode<S> paramNode = root;
-            CommandParameter[] parameters = rootCommand.getDefaultRegisteredCommand().parameters;
-            registerParameters(parameters, rootCommand, paramNode, suggestionProvider, executor, permCheckerRoot);
+        RegisteredCommand defaultCommand = rootCommand.getDefaultRegisteredCommand();
+        if (defaultCommand != null) {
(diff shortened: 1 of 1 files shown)


JOO200:
Maybe extract the duplicated code in a new method?

brigadier/src/main/java/co.aikar.commands/ACFBrigadierManager.java
```diff
@@ -89,6 +89,12 @@
         root = rootBuilder.build();
         boolean isForwardingCommand = rootCommand.getDefCommand() instanceof ForwardingCommand;
 
+        if (rootCommand.getDefaultRegisteredCommand() != null) {
+            CommandNode<S> paramNode = root;
+            CommandParameter[] parameters = rootCommand.getDefaultRegisteredCommand().parameters;
+            registerParameters(parameters, rootCommand, paramNode, suggestionProvider, executor, permCheckerRoot);
```
JOO200, line 95: Why do you pass the rootCommand and permCheckerRoot here? pretty sure it's enough to pass `rootCommand.getDefaultRegisteredCommand()` and `permCheckerSub`.

Your current implementation will provide wrong suggestions for @Default command methods with a @CommandPermission.

brigadier/src/main/java/co.aikar.commands/ACFBrigadierManager.java
```diff
@@ -161,4 +149,31 @@
         return root;
     }
 
+    <C> void registerParameters(CommandParameter[] parameters,
```
JOO200, line 152: Doesn't need to be a generic method when you're only using RegisteredCommand and the permCheckerSub

JOO200 (changes requested):

DeathRealms:
I still need to figure out why when registering the parameters for the default method it prevents sub commands you don't have permission for from being removed from tab completion, but they get removed if you don't have permission for both the default and sub command.

chickeneer:
This change looks reasonable. In regards to the latter comment about 

> when registering the parameters for the default method it prevents sub commands you don't have permission for from being removed from tab completion, but they get removed if you don't have permission for both the default and sub command.

I think it clear you are making this comment in general and shouldn't block merging this.

Issue #296, opened 2020-10-04
Bukkit help doesn't get translated


RoinujNosde:
Running `/bukkit:help SimpleClans`, where SimpleClans is the plugin name
![bukkit help](https://user-images.githubusercontent.com/6441328/95024184-e216d280-0657-11eb-816a-eb6a175cba6f.PNG)
Running `/clan acc`, where `clan` is registered via ACF
![acf help](https://user-images.githubusercontent.com/6441328/95024187-e4792c80-0657-11eb-9567-47528de02087.PNG)

As you can see, on ACF's help the message is translated normally.

Command code:
```java
    @CommandAlias("%accept")
    @Description("{@@command.description.accept}")
    @Conditions("can_vote")
    public void accept(Player player, ClanPlayer cp) {
        Clan clan = cp.getClan();
        if (clan != null) {
            clan.leaderAnnounce(GREEN + lang("voted.to.accept", player.getName()));
        }
        requestManager.accept(cp);
    }
```

PR #358, opened 2022-01-30
Fixes /help <plugin> messages not localized #296

RoinujNosde:
Before:
image

After:
image


chickeneer:
Thanks for the help on this. 👍 

General note on this PR in case we look back on this. getDescription() needs to return a normal readable String as it is overriden from the Bukkit getDescription, so non-acf Bukkit help calls that method to grab descriptions. 
This change should not have broken anything else as anything acf internal should use the description/helptext fields more directly and then formatMessage to translate.

PR #360, opened 2022-02-20
Fix for servers that extended the ServerPlayer class, and thus locale…

ChromMob:
… is no longer a declared field


RoinujNosde:
Duplicate code
Could be prevented by adding a method like:
```java
getField(clazz, "locale"):
    try {
        return clazz.getDeclaredField
    } catch {
        return clazz.getField
    }
```

JOO200:
And it might be easier to just override the method in PaperCommandManager and use player$getLocale() instead.
Unless your server fork is not a paper fork and a old version

PR #381, opened 2023-01-05
Add allow illegal name manager option

AlexProgrammerDE:
Some servers may have players with "invalid" names, such as geyser with a name prefix.

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/ACFBukkitUtil.java
+++ b/bukkit/src/main/java/co/aikar/commands/ACFBukkitUtil.java
@@ -282,7 +282,7 @@ public class ACFBukkitUtil {
 
         //noinspection Duplicates
         if (matches.isEmpty()) {
-            if (!isValidName(name)) {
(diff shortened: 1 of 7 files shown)


bukkit/src/main/java/co/aikar/commands/ACFBukkitUtil.java
```diff
@@ -319,8 +319,8 @@ private static void findMatches(String search, CommandSender requester, List<Pla
     }
 
 
-    public static boolean isValidName(String name) {
-        return name != null && !name.isEmpty() && ACFPatterns.VALID_NAME_PATTERN.matcher(name).matches();
+    public static boolean isValidName(String name, CommandManager manager) {
```
JOO200, line 322: This is public therefore public api. There should be a backwards compatible method for those who use that method in their plugins.

AlexProgrammerDE: Done!

PR #382, opened 2023-02-21
Allow illegal name

AlexProgrammerDE:
Continuation of #381 as I changed the base branch

Previous message:
Some servers may have players with "invalid" names, such as geyser with a name prefix.


chickeneer:
Merged this PR with an addition ^ see tagged commit.

Here was my commit comment.

I see the value in allowing acf users to disable the valid name check.
I think others may want to go a step further and create their own
valid name verification. Those that want to truly accept all names can simply
do `CommandManager#setValidNamePredicate(name -> true)` with the new API.
Another solution I considered was having users override the isValidName method
on their CommandManager, but this functionality seemed more friendly.

PR #389, opened 2023-04-16
Update BukkitCommandContexts.java

casperwtf:
The point of this is to add better support for geyser since it will first check if it can match an online player with that name.

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
+++ b/bukkit/src/main/java/co/aikar/commands/BukkitCommandContexts.java
@@ -140,9 +140,15 @@ public class BukkitCommandContexts extends CommandContexts<BukkitCommandExecutio
                     throw new InvalidCommandArgument(MinecraftMessageKeys.NO_PLAYER_FOUND_OFFLINE,
                             "{search}", name);
                 }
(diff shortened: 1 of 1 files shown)


chickeneer:
I have never used geyser. But I can't figure out why this would improve geyser support.

I am looking at getOfflinePlayer() and it already calls getPlayer() as one of the first thing it does. The exception is the string based version uses getPlayerExact - not the getPlayer variation.

I think I am lacking information to make this change.

sqlerrorthing (approved):

chickeneer:
Hey there. I am closing this PR. As previously mentioned - I need some guidance on why this functionally would do anything different than what already happens.

the getOfflinePlayer#UUID method definitely does nothing different as I see it.
but would be interested to know why getPlayerExact isn't getting what you need.

sqlerrorthing:
oh :(

chickeneer:
If someone wanted to explain elsewhere. They could explain why this change is needed in the discord as well.
https://aikardiscord.emc.gs/

If someone gave a reason, we could still pull it. Just I don't see a reason for it to sit open at this point as the change doesn't appear to do anything to me.

PR #411, opened 2024-10-24
Update/sponge

Intybyte:
Update sponge api to 10.1.0 from 5.1.0

Needs testing and reviewing


Intybyte:
Testing on 1.19.4 succesfull, the way sponge handles command completition is weird to me, for example having a `/test` command and defining a `string` subcommand will allow you to tab `/test string string string` which might not be intended behaviour

Intybyte:
A testable example here https://github.com/Intybyte/acf-testing-sponge

chickeneer:
I have merged this PR. It passes the cursory review. As an initial version for this updated sponge version. It compiles, and should work for the most part. Getting it merged and deployed is a good first step for if we need to make functionality improvements.

PR #417, opened 2025-02-02
No longer suggest input as tab completion

InstantlyMoist:
This PR fixes ACF giving the players input back as a tab-complete suggestion.

Previously:
image
Expected was to only receive 'skull' right here, as that's the only registered `@Subcommand`

After:
image
No suggestion is given, and it auto-fills to the correct subcommand.

I'm not entirely sure what the impact of this is, as I only started using the framework 2 days ago. But I'm more than happy to receive feedback.

### DIFF
--- a/paper/src/main/java/co/aikar/commands/PaperAsyncTabCompleteHandler.java
+++ b/paper/src/main/java/co/aikar/commands/PaperAsyncTabCompleteHandler.java
@@ -77,6 +77,7 @@ class PaperAsyncTabCompleteHandler implements Listener {
 
         BukkitCommandIssuer issuer = this.manager.getCommandIssuer(sender);
         List<String> completions = rootCommand.getTabCompletions(issuer, commandLabel, args, false, async);
+        completions.removeAll(Arrays.asList(args));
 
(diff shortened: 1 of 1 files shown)


chickeneer:
Hey there. I have looked into this. This might fix your specific example, but I do think it is a solution that could have unintended consequences. Let's say if the typed in text is already valid text.

So, my big question was why the `s` was getting added to the completions in the first place. So code searching led me to the two following code spots.

https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/CommandCompletions.java#L216

https://github.com/aikar/commands/blob/master/core/src/main/java/co/aikar/commands/CommandCompletions.java#L305

I will have to do some tinkering to figure out which can be axed safely. On one hand, an empty list could be safe to tell the client just that nothing was found. But I want to make sure that there are no side effects like Tab-ing deleting deleting what they had already typed in for some reason.

I will look into this further. Leaving this open for now. Again, I don't think this is the desired solution.

chickeneer:
As previously indicated. I 'think' that the PR'd solution is likely too aggressive. I have attempted a change. As indicated in Discord - I don't have an easy way to test a 1.12 server - but I hope this resolves the problem you are experiencing.

PR #418, opened 2025-02-08
Feature/pipe or permissions

Intybyte:


chickeneer:
I merged this branch. And then immediately added in an extremely simple operator precedence between the AND and OR (commas and pipes respectively). https://github.com/aikar/commands/commit/6629dc2f46e6f10fe3b10c65dc154c10e4ea506c

A general note to any random readers of this PR. It is not an objective of this project to define anything more complex with the parsing system for permission checking. ACF users could design/implement their own to fit their individual needs.

PR #421, opened 2025-03-04
Update/sponge 8 fix

Intybyte:
This fix allows compatibility for sponge8 till sponge 10 (tested at least) and is more flexible towards sponge api breakages.

Requires testing (will test in a few days or when am free, or someone else ig).


chickeneer:
I am concerned about the removal of the removal of constructors for SpongeCommandSource. Admittedly, I have no way to know if anyone is or isn't using it. But if I simply merge and deploy this to the current version of ACF - it has the technical potential to break if for some reason someone had called that API.

I am aware that within the ACF module - it doesn't have an affect, but would prefer to find a solution which keeps constructors where possible with a Deprecated or something.

chickeneer:
I should say. I am not necessarily opposed to merging it as proposed. Just want the before mentioned risk be part of the conversation.

PR #422, opened 2025-03-08
Open ACFPatterns class to be public

benwoo1110:
While coding my plugin, I find myself having to re-compile many of the patterns that is already provided by ACFPatterns. Thus I think it makes sense to make the class public for usage outside just ACF.


chickeneer:
I understand why this change may be convenient. But I would not label this pattern class as API of the ACF library (nor desire it to be). By making it public, it adds a layer of immutability to it - where there would not be the freedom to refactor this utility class for the project as needed. Removing unused patterns or even changing the patterns themselves. Because there will now be the chance that someone is using it in a plugin.

Maybe if we incremented ACF versions more often, I would be more open to it becoming an API. I can't imagine that the data size of this class is that large, that it would be overly burdensome to implement your own patterns. They are relatively basic.

PR #423, opened 2025-03-08
Use PlayerLocaleChangeEvent instead of repeating task when available

benwoo1110:
Use PlayerLocaleChangeEvent for newer versions of minecraft instead of polling all players every few second to see if their locale has changed. This will be a slight performance improvement as player usually dont change locale that often and will very inefficient to keep polling. The polling method will still be kept as a fallback for older minecraft version where PlayerLocaleChangeEvent is not present.

### DIFF
--- a/bukkit/src/main/java/co/aikar/commands/ACFBukkitLocalesListener.java
+++ b/bukkit/src/main/java/co/aikar/commands/ACFBukkitLocalesListener.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2016-2025 Daniel Ennis (Aikar) - MIT License
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining
+ *  a copy of this software and associated documentation files (the
+ *  "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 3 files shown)


chickeneer:
I agree with the goal of this PR. I think it is going to fall a tad short currently. Unless I am wrong (totally possible). I don't believe that this will pull the joining player's original locale. Only if they change it after connecting.

So minimally there would need to be a call for every joined player locale once. And then this event to check if they have changed locales.

Seems like this could probably be accomplished just through a PlayerJoinEvent listener. But not sure without doing some testing myself.

Hmm.

benwoo1110:
There is already existing code that does it in PlayerJoinEvent

benwoo1110:
Somewhere here is the relevant code: https://github.com/aikar/commands/blob/master/bukkit/src/main/java/co/aikar/commands/ACFBukkitListener.java#L57

chickeneer:
ahah! Indeed I see that now. Thank you. I will double check things. I think resolves the only red flag that went through my brain then.

chickeneer:
Yup. All looks good. Merged in. Deploying out now.

PR #429, opened 2025-04-06
Add support for getting annotation from parent class/method/parameter

benwoo1110:
This will allow use cases such as having a parent class with `@CommandAlias` so I do not have to add `@CommandAlias("mv")` to 20 different command classes.
```java
@CommandAlias("mv")
public abstract class CoreCommand extends MultiverseCommand {
    // ...
}

class CloneCommand extends CoreCommand {
    @Subcommand("clone")
    void onCloneCommand(...) { }
}
```

And this will also me to create aliases to a command by extending the command class as such:
```java
class LegacyAlias extends CloneCommand {
    @Override
    @CommandAlias("mvclone")
    void onCloneCommand(...) {
        super.onCloneCommand(...);
    }
}
```

### DIFF
--- a/core/src/main/java/co/aikar/commands/Annotations.java
+++ b/core/src/main/java/co/aikar/commands/Annotations.java
@@ -60,7 +60,7 @@ class Annotations<M extends CommandManager> extends AnnotationLookups {
             } else if (object instanceof Method) {
                 annotation = getAnnotationFromParentMethods((Method) object, annoClass);
             } else if (object instanceof Parameter) {
-                annotation = getAnnotationFromParameters((Parameter) object, annoClass);
+                annotation = getAnnotationFromParentParameters((Parameter) object, annoClass);
             }
         }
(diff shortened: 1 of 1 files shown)


chickeneer:
This seems eligible for merge. First, I am trying to just contemplate if there are any negative consequences to a "parent" class/method/parameter has an annotation that we don't want on the child.

I cannot immediately think of any such situation. Just seems like this opens the opportunity for a new 'issue' of being unable to unmark an annotation.

I don't think this changes behavior so it is likely safe on that front. And users can just design with this limitation in mind.

benwoo1110:
If really the need to customise, we can consider opening the getAnnotation method in command manager to be overridable, and we can implement our own class that extends AnnotationLookup.

chickeneer:
This might be fine. Just tryping out my thoughts. Just don't want to pull it and then second guess it later

PR #440, opened 2026-05-04
Add support for having custom completion filters

benwoo1110:
Previously, all tab completions were filtered based on "starts with", while this is ok most of the time, it doesn't really work too well when there is a long list of completions, and it's hard to remember what the starting word is. For example, completion such as `enforce-gamemode` if I can't quite remember and thought it's`gamemode-enforce` and type `gamemode`, no completion will be suggested. Another problem is with Minecraft's increased use of NamespaceKey, and "starts with" tab complete will mean users need to type in `minecraft:<key>` just to find the item instead of typing <key> and letting the tab complete suggest the full NamespaceKey.

This PR adds a new `CommandCompletionFilter` to solve the issue. This allows each completion handler to set its individual filter type based on what is best for the data it's presenting, while preserving the default behaviour of "starts with" filtering.

### Using Contains
```java
completions.registerStaticCompletion("mvconfigs", config.getStringPropertyHandle().getAllPropertyNames(), CommandCompletionFilter.contains());
```
Screenshot 2026-05-05 at 12 07 46 AM

### Custom Namespace Filtering
```java
private static final CommandCompletionFilter<BukkitCommandCompletionContext> NAMESPACEDKEY_FILTER = (context, completion) -> {
    String[] split = completion.split(":", 2);
    if (split.length < 2) {
        return ApacheCommonsLangUtil.startsWithIgnoreCase(completion, context.getInput());
    }
    return ApacheCommonsLangUtil.startsWithIgnoreCase(split[0], context.getInput()) ||
            ApacheCommonsLangUtil.startsWithIgnoreCase(split[1], context.getInput());
};


completions.registerAsyncCompletion("mvworlds", this::suggestMVWorlds, NAMESPACEDKEY_FILTER);
```
image

### DIFF
--- a/core/src/main/java/co/aikar/commands/BaseCommand.java
+++ b/core/src/main/java/co/aikar/commands/BaseCommand.java
@@ -651,7 +651,7 @@ public abstract class BaseCommand {
                 }
             }
 
-            return filterTabComplete(args[args.length - 1], cmds);
+            return cmds;
         } finally {
             postCommandOperation();
         }
@@ -706,8 +706,7 @@ public abstract class BaseCommand {
             return Collections.emptyList();
         }
 
-        List<String> cmds = manager.getCommandCompletions().of(cmd, issuer, args, isAsync);
-        return filterTabComplete(args[args.length - 1], cmds);
+        return manager.getCommandCompletions().of(cmd, issuer, args, isAsync);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


chickeneer:
A couple minute look, and this looks good. I need to load it up in my IDE to sanity check it. But should be able to merge it later today.

benwoo1110:
Alright, lmk if there is anything I need to change/fix.

chickeneer:
I am going ahead and merging and deploying this. I think this looks really solid across the board.
My biggest concern was if there was any change of behavior for existing api usage, etc. And I really don't think there is regression in that regard.

With that said, in the next week - I still want to do some testing on the extent of these new features. Not that I think anything is wrong - but I think you introduced some cool options - that I want to think about more. I like it. :)

